Issue #125, opened 2015-04-30
provisioner error when spinning up vagrant based instances


asnaedae:
Following error thrown when running `sh bootstrap/apollo-launch.sh`:

```
 There are errors in the configuration of this machine. Please fix
the following errors and try again:

vm:
* The '' provisioner could not be found.
```

PR #127, opened 2015-04-30
PR issue #125 failure provision on vagrant

asnaedae:
update docs to require vagrant-hosts plugin install, and throw error should plugin not be detected.


tayzlor:
Merged

Issue #54, opened 2015-03-31
Create a drupal demonstrator


tayzlor:
If we can use the michaelpage site - that would be nice

Basically mysql / memcache / drupal in containers that we can deploy to the stack. Depends on #53

tayzlor:
Some work has begun on this over here https://github.com/Capgemini/apollo-drupal-commerce-kickstart

PR #139, opened 2015-05-06
Addresses #54 - initial commit add support for deploying drupal on marathon

tayzlor:
Opening up for initial feedback.

### DIFF
--- a/examples/drupal-mysql/README.md
+++ b/examples/drupal-mysql/README.md
@@ -1,0 +1,57 @@
+## Drupal / MySQL example
+
+This example shows how to deploy a multi-tier Drupal web app using Apollo and Marathon Mesos framework.
+
+The example includes -
+
+* A web frontend (Using Drupal Commerce Kickstart distribution)
+* A MySQL (5.5) backend
+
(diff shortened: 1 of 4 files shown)


examples/drupal-mysql/README.md
```diff
@@ -0,0 +1,60 @@
+## Drupal / MySQL example
```
enxebre: Shall we create a group json marathon file with the drupal service being a dependency of the mysql one so just one step for deploying?

examples/drupal-mysql/README.md
```diff
@@ -0,0 +1,57 @@
...
+First, edit ```drupal-mysql.json```, in the mysql app service definition, to use database credentials that you specify (or leave as is if you are happy and this is just for demo purposes).
+
+If you changed the database credentials as above, then edit the drupal app definition in ```drupal-mysql.json```, to use database credentials that you specify (or leave as is if you are happy and this is just for demo purposes).
+
+Start the services by running the following command (from somewhere you can access the REST interface of Marathon). Replace $MARATHON_IP with the actual hostname/ip address of the marathon instance.
+Marathon by default should be deployed in the ```mesos_master``` nodes.
+
+```
+curl -X POST -HContent-Type:application/json -d @drupal-mysql.json $MARATHON_IP:8080/v2/groups
+```
```
enxebre: We could include a comment on how to delete it -> curl -XDELETE  $MARATHON_IP:8080/v2/groups/drupal-mysql-example

PR #142, opened 2015-05-07
cleanup of markdown for the digitalocean instructions

asnaedae:
extra code escape causing markdown to render badly, also changed a few words to remain consistent across the docs


tayzlor:
Thanks!

Issue #133, opened 2015-05-05
Allow override of mesos task port range


tayzlor:
By default tasks / containers start up with port range starting at 31000 > 

We should make this overridable via ansbile

PR #143, opened 2015-05-07
Allow override of docker options for weave

tayzlor:


tayzlor:
Meh accidentally added fixes for #133 into this PR.

tayzlor:
Ah looks like I broke a few tests, fixing now

PR #150, opened 2015-05-08
[terraform/aws] nat uses atalas artifact for ami

pmbauer:
- previously, nat-server.tf used the default eu-west-1 ami
- this is not portable when spinning up infrastructure in any other
  region
- parameterize the nat ami similar to master/slave with an atlas
  artifact
- bubble up this parameter through the bootstrap/aws scripts


tayzlor:
Thanks for the PR, however i'm not sure its quite the correct solution. The NAT server doesn't really need to use the baked packer image because we are only putting on it docker + ovpn docker container. For this reason we just used a stock ubuntu AMI (from the eu-west1 region). 

Your PR however does highlight a couple of issues at play here - 

- The packer image (for the AMI) has only been published to the eu-west1 region (https://github.com/Capgemini/Apollo/blob/master/packer/ubuntu-14.04_amd64-amis.json#L23). We should probably fix that so its available in other regions. 
- The nat server is only using a stock ubuntu from the eu-west-1 region (https://github.com/Capgemini/Apollo/blob/master/terraform/aws/variables.tf#L76-L82) . That should be expanded to support more than one reason.

I've opened #153 related to this

pmbauer:
Thanks; that's perfectly okay. I'll close this then.

One problem with the nat-server.tf as is if you install docker (curl -sSL https://get.docker.com/ubuntu/ | sudo sh), you need a sleep prior to that similar to https://github.com/Capgemini/Apollo/blob/master/packer/scripts/ubuntu/base.sh#L9 since the docker install script does an apt-get update.  I've seen it fail without the sleep.

PR #151, opened 2015-05-08
[roles/mesos] parameterize slave work_dir

pmbauer:
- on aws, the large ephemeral volume is mounted at /mnt by default
- without setting the slave's working dir to /mnt/X, the very small ebs
  boot volume will be filled up VERY quickly making the cluster
  unoperable.
- bubble up this parameter to the bootstrap/aws scripts


tayzlor:
Thanks!

pmbauer:
thanks!

Issue #152, opened 2015-05-08
Handle environment variables more dynamically


tayzlor:
**Problem** 

Every time someone wants to override something in the bootstrap scripts this involves placing the environment variable in ```config-default.sh``` and modifying ```--extra-vars``` for Ansible. 
We have a similar issue for terraform where we need to modify ```util.sh``` to add ```-var ...``` on to the terraform plan.
This is not really scalable.

**Proposal**

Part of the solution is to move to using terraform environment variables (introduced in terraform 0.5). See #148

The other idea is to handle any environment variables by doing something like the following - 

- namespace all environment variables ```APOLLO_``` 
- on bootstrap loop all ```APOLLO_``` env variables and build up --extra-vars from those

e.g. 

```APOLLO_CONSUL_DC=eu-west-1 ``` would become ```--extra-vars "consul_dc=eu-west-1"```

This means when we add new features we do not need to surface any extra code in the bootstrap scripts

enxebre:
I like how we solve this in #138 for ansible. I would keep that (even if we split the PR )

For terraform even if we do the same approach and we allow to override via "APOLLO_", then "TF_VAR_name" will still be avalable once we move to 0.5.0 so they are compatible. 

If we follow "APOLLO_" approach we need do differentiate some how between ansible and terraform otherwise we could end up with variables overlapping between the tools. So we would nee to do something like "APOLLO_TF" which makes me think why dont we just use "TF_VAR_name". So cant see a real benefit for using custom approach for terraform right now appart from cross tool project consistency. I think terraform question will be answered itself once the apollo plugin concept for terraform becomes more clear.  (see #140)

tayzlor:
for terraform - I would agree - lets just use TF_VAR_ - no point adding an additional layer here that does nothing.

Issue #148, opened 2015-05-08
Allow toggling of debug output via a single switch (env var)


tayzlor:
We should be able to toggle ansible / terraform verbosity logs via a single debug switch in.

so turning on verbose would do ```TERRAFORM_LOG=1``` and run ansible with ```-vvv```

asnaedae:
Would the preference be to set via environment variable (which would follow convention with cloud provisioning patterns) or an option to apollo-launch.sh?

tayzlor:
I think my personal preference would be to set a single environment variable (APOLLO_LOG??), which would in turn set -vvv  on ansible and TERRAFORM_LOG=1

asnaedae:
It's what I was thinking would match with  existing usage of of the APOLLO_ namespace, will put together a PR

tayzlor:
Only thing we need to consider is how this plays with https://github.com/Capgemini/Apollo/issues/152

PR #155, opened 2015-05-11
issue #148 toggling of debug

asnaedae:
left it pretty lean so to allow for refactoring based on #152 - only tested against vagrant thus far.


bootstrap/apollo-env.sh
```diff
@@ -14,3 +16,10 @@ if [[ -z "${color_start-}" ]]; then
   declare -r color_norm="${color_start}0m"
 fi
```
enxebre: What do you think about providing an option for ensuring they are disable too like:
```
if [[ "${APOLLO_LOG}" == "0" ]]; then
unset TF_LOG
unset ANSIBLE_LOG
fi
```
Should we also provide a level system so:
APOLLO_LOG=1 -> -v
APOLLO_LOG=2 -> -vv
APOLLO_LOG=3 -> -vvv
APOLLO_LOG=4 -> -vvvv

Then if we unset whatever there is in TF_LOG and ANSIBLE_LOG Applies

bootstrap/apollo-env.sh
```diff
@@ -14,3 +16,10 @@ if [[ -z "${color_start-}" ]]; then
...
 
+# change logging levels of called components at a global level
+# if unset allow for existing selective logging of components
+if [[ "${APOLLO_LOG}" != "error" ]]; then
+  #export VAGRANT_LOG="debug"
```
enxebre: Are we including or can just remove this one?

bootstrap/apollo-env.sh
tayzlor: I dont personally think we need to make it as complicated as a level thing (3 -vvv would be fine) - but i'm happy to be put in my place if you have a good argument for it.

enxebre: The levels just gives you a centralized way to manage different combinations. The way it is right now once I set APOLLO_LOG there is no way to silence the Apollo log but disabling each indivdually (2 at the moment, we might want add more like packer/vagrant). So I'd definitely like an APOLLO_LOG=0
Then if you unset APOLLO_LOG you can set differents levels via ANSIBLE_LOG or per tool.

asnaedae: That sounds sensible, I didn't want to overwrite any previously/externally defined logging as you're quite often only interested in a particular application.

bootstrap/apollo-env.sh
asnaedae: I'll remove it from the code

bootstrap/apollo-env.sh
asnaedae: How about we use the numeric as options, however only cater for 0 and N - then as we gain information to which log level combinations are useful create canned groups for them via APOLLO_LOG

enxebre: sounds good to me

bootstrap/aws/util.sh
```diff
@@ -10,10 +10,6 @@ verify_prereqs() {
     echo -e "${color_red}Can't find terraform in PATH, please fix and retry.${color_norm}"
     exit 1
   fi
```
enxebre: think you need to rebase on to master

asnaedae: Had thought that's what I did already... will resolve.

bootstrap/aws/util.sh
```diff
@@ -75,7 +75,7 @@ ansible_playbook_run() {
     ANSIBLE_SSH_ARGS="-o UserKnownHostsFile=/dev/null -o ControlMaster=auto -o \
     ControlPersist=60s -F $APOLLO_ROOT/terraform/aws/ssh.config -q" \
     ansible-playbook --user=ubuntu --inventory-file=$APOLLO_ROOT/inventory/aws \
-    --extra-vars "consul_atlas_infrastructure=${ATLAS_INFRASTRUCTURE} \
+    ${ANSIBLE_LOG} --extra-vars "consul_atlas_infrastructure=${ATLAS_INFRASTRUCTURE} \
```
tayzlor: if we fall into the ```error)``` case above and nothing is set for ```ANSIBLE_LOG``` will we get a warning here because the variable is undefined?

tayzlor:
code LGTM. Wondering if we just need a small doc entry for this to explain how to turn on/off debug mode

bootstrap/apollo-env.sh
```diff
@@ -14,3 +17,24 @@ if [[ -z "${color_start-}" ]]; then
...
+    echo "Forcing reduction of component logging"
+    unset TF_LOG
+  ;;
+  1)
+    export TF_LOG=1
+    export ANSIBLE_LOG="-v"
+  ;;
+  *)
+    export TF_LOG=${APOLLO_LOG}
+    export ANSIBLE_LOG="-vvvv"
```
enxebre: Not sure this is gonna work for vagrant as it is expecting 'vvvv' no '-vvvv' see http://docs.vagrantup.com/v2/provisioning/ansible.html

roles/serverspec/tasks/main.yml
```diff
@@ -18,7 +18,6 @@
     - serverspec
 
 - name: run serverspec tests
-  sudo: yes
```
tayzlor: the sudo needs to stay here, otherwise we fail on testing docker containers

enxebre:
Closing in favour of https://github.com/Capgemini/Apollo/pull/391

Issue #160, opened 2015-05-11
Upgrade mesos version in the packer box to 0.22.1


tayzlor:
We should do this at the same time we fix out versioning for the boxes via #159

tayzlor:
Fixed via https://github.com/Capgemini/Apollo/commit/3798eabbf1dbef7b985817adfe5642cc97d9bf73 we just need to rebuild the packer images.

PR #162, opened 2015-05-11
Fix #159 - Packer box versioning

### LINKED ISSUE #159: Come up with a sane way of managing vagrant box versioning
**Problem** 

When we initially released the vagrant box we versioned it with the current version of mesos (0.22.0). 

That doesn't really make much sense though, because we probably want to release versions of the box outside of the mesos release cycle :)

We need to come up with a sane way (semantic versioning?) of keeping track of the box versions and tracking those through release branches.

tayzlor:
This should also tackle #160

approach i've taken follows http://semver.org/

>Build metadata MAY be denoted by appending a plus sign and a series of dot separated identifiers immediately following the patch or pre-release version. Identifiers MUST comprise only ASCII alphanumerics and hyphen [0-9A-Za-z-]. Identifiers MUST NOT be empty. Build metadata SHOULD be ignored when determining version precedence. Thus two versions that differ only in the build metadata, have the same precedence. Examples: 1.0.0-alpha+001, 1.0.0+20130313144700, 1.0.0-beta+exp.sha.5114f85.

### DIFF
--- a/packer/ubuntu-14.04_amd64-google.json
+++ b/packer/ubuntu-14.04_amd64-google.json
@@ -2,10 +2,13 @@
   "variables": {
     "account_file": "{{env `GCS_ACCOUNT_FILE`}}",
     "project_id": "{{env `GCS_PROJECT_ID`}}",
-    "mesos_version": "0.22.0-1.0.ubuntu1404",
+    "source_image": "{{env `GCS_SOURCE_IMAGE`}}",
+    "zone": "{{env `GCS_ZONE`}}",
+    "mesos_version": "0.22.1-1.0.ubuntu1404",
     "marathon_version": "0.8.1-1.0.171.ubuntu1404",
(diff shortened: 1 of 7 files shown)


packer/ubuntu-14.04_amd64-google.json
```diff
@@ -15,8 +18,8 @@
     "ssh_timeout": "10m",
     "account_file": "{{user `account_file`}}",
     "project_id": "{{user `project_id`}}",
-    "source_image": "ubuntu-1404-trusty-v20150316",
-    "zone": "europe-west1-b"
+    "source_image": "{{user `source_image`}}",
+    "zone": "{{user `zone`}}",
```
wallies: no comma needed on end here

tayzlor:
Created #163 as a follow-up

Issue #101, opened 2015-04-23
Move serverspec testing to be via ansible


tayzlor:
See these repos - 

https://github.com/volanja/ansible-sample-tdd
https://github.com/volanja/ansible_spec

enxebre:
moving this to critical as we are not running serverspecs at all after provision since we moved to ansible

wfindley:
Server spec tests?  I love you.

asnaedae:
we love serverspec too!

PR #169, opened 2015-05-12
Initial commit of serverspec test harness

tayzlor:
This is a start towards #101. Opening up for initial review, with some caveats -

- Not been developed/tested yet on Digitalocean
- Not been developed/tested yet on AWS

**Outstanding questions/thinking**

- Should we run serverspecs by default, or have them off and allow them to be run by toggling some variable?
- ```tests/properties.yml``` references the hostnames > specs we need to test. Our hostnames are not consistent across providers (vagrant/do/aws) - we either need to make them consistent, or handle it otherwise
- We probably need to increase the test coverage (but this could be done afterwards)
- On vagrant the slave test fails (the slave service is not running correctly - this is a bug in the vagrant provisioning) - created #172 
- On vagrant the master test fails to see the marathon service - I believe this is because there are 2 startup scripts for marathon (one via upstart and one via init.d) and it was looking at the init.d one, which was stopped - so we might need to alter that test or use the init.d startup script - both appear to work fine (I manually started up with both scripts and they both worked)


tests/test.yml
```diff
@@ -0,0 +1,20 @@
...
+      command: gem install bundler --no-ri --no-rdoc
+      args:
+        creates: /usr/local/bin/bundler
+    - name: install bundle files
+      command: bundle install --path vendor
+      args:
+        chdir: "{{ tests_path }}"
+        creates: "{{ tests_path }}/vendor"
+    - name: run serverspec tests
+      command: "bundle exec rake serverspec:{{ ansible_hostname }}"
```
enxebre: can we just use tags instead of hostname so it would be consistent across envs

tests/properties.yml
```diff
@@ -0,0 +1,24 @@
+master1:
```
enxebre: If we use tags instead hostname (see comment below) we could do dry here and just have:
```
+master_tag:
+  :roles:
+    - dnsmasq
+    - mesos_master
+    - marathon
+    - zookeeper
+slave_tag:
+  :roles:
+    - dnsmasq
+    - mesos_slave
+    - weave
+    - docker
```

wallies:
think serverspecs should be a variable and we use tags like alberto mentioned above. fixed marathon spec by running it under upstart.

wallies:
#172 is fixed now, so is the marathon service. So just need to fix the tags

tests/spec/marathon/marathon_spec.rb
```diff
@@ -0,0 +1,13 @@
...
+
+describe package('marathon') do
+  it { should be_installed }
+end
+
+describe service('marathon') do
+  it { should be_running.under('upstart') }
+end
+
+#describe port(8080) do
```
tayzlor: This test was failing

.gitignore
```diff
@@ -9,3 +9,6 @@ packer/build
 .wercker
 *.cache
 ssh.config
+vendor
+.bundle
+.kitchen
```
enxebre: Do we need .kitchen?

roles/serverspec/defaults/main.yml
```diff
@@ -0,0 +1,5 @@
+---
+# defaults file for serverspec
+serverspec_run_tests: false
+serverspec_tests_path: /vagrant/tests
```
enxebre: Do we need to be committed to vagrant in the naming?

.gitignore
tayzlor: i was experimenting with it and added it, no harm to include it anyway

roles/serverspec/defaults/main.yml
tayzlor: this is just the default value - can be overridden elsewhere

Issue #167, opened 2015-05-12
Occassionally mesos isnt able to send registration acknowledgement to marathon


tayzlor:
Tested in digitalocean spinning up a fresh cluster - 

Looking at the mesos logs we see lots of 

```
I0512 06:10:11.741300  2039 master.cpp:1574] Received registration request for framework 'marathon' at scheduler-3d379f04-cb71-44bf-bd05-2194d9c3f283@127.0.1.1:51954
I0512 06:10:11.741420  2039 master.cpp:1625] Framework 20150512-055934-1814717742-5050-2011-0000 (marathon) at scheduler-3d379f04-cb71-44bf-bd05-2194d9c3f283@127.0.1.1:51954 already registered, resending acknowledgement
```

doing a -

```
restart marathon
```

seems to bring the cluster back into sync, so there's obviously something going amiss on initial deploy.

wallies:
cannot reproduce after bringing up fresh cluster 3 times.

tayzlor:
Closing via #184

PR #184, opened 2015-05-13
Fix #175 - add tags for roles, and some slight refactoring / style fixes

### LINKED ISSUE #175: Tag all tasks in ansible roles
Not all our tasks are tagging in the ansible roles. We should add tags to everything so it makes it easier to re-run playbooks with only specific tags using the ```--tags=``` command line flag

tayzlor:


tayzlor:
This also fixes a bug where the consul datacenter is not populated correctly.
Additionally with this branch I cannot reproduce #167

PR #185, opened 2015-05-13
Rename packer / artifacts to just apollo, rather than apollo-mesos

tayzlor:
BIG FAT BREAKING CHANGE.

We need to rebuild all artifacts in correlation with this PR as it changes ```apollo-mesos``` to just ```apollo``` in most of those.

Suggest we keep the artifacts at apollo-mesos in the vagrantcloud and atlas anyway for a bit so we dont break people on 0.1


tayzlor:
don't merge this until we've sorted out the artifacts in atlas. Atlas is a bit flakey at the moment :/

asnaedae:
This runs foul of #197

Issue #113, opened 2015-04-28
Create an upstart script for weave


tayzlor:
We have a bit of a convoluted ansible task for configuring weave. This could be simplified if we just provisioned an upstart script and started up the weave service (and let the script handle the dependencies). On a different project (which was using puppet) we had this - 

```
description "Weave Docker Network Service"
start on filesystem and started docker
stop on runlevel [!2345]

script
 [ -e /etc/default/weave ] && . /etc/default/weave
 /usr/bin/docker rm -f weave || true
 <%= scope.lookupvar('weave::bin_dir') %>/weave launch <% if @cluster_peers %><%= @cluster_peers %><% end %>
end script

respawn
respawn limit 10 10
kill timeout 10
```

We should be able to achieve something similar here. See also http://docs.docker.com/articles/host_integration/

tayzlor:
Bumping this in priority since this could make spinning up weave in the host a bit less prone to random provisioning errors

tayzlor:
Putting a note in here around trying to use weave connect instead of launching with peers to try and avoid random timeouts communicating between machines that result in - 

```
fatal: [46.101.42.108] => One or more undefined variables: 'dict object' has no attribute 'ansible_eth0'
```

tayzlor:
Can we also make this dependent on docker. So instead of having to do - 

```

- name: configure weave bridge for docker
  sudo: yes
  lineinfile: "dest=/etc/default/docker state=present regexp=^DOCKER_OPTS= line='DOCKER_OPTS=\"{{ weave_docker_opts }}\"'"
  notify:
    - Restart docker
    - Weave launch
```

We could just do 

```

- name: configure weave bridge for docker
  sudo: yes
  lineinfile: "dest=/etc/default/docker state=present regexp=^DOCKER_OPTS= line='DOCKER_OPTS=\"{{ weave_docker_opts }}\"'"
  notify:
    - Restart docker
```

the restart docker would trigger a stop / start of weave.

We might have to add to the init script ```stop on stopping docker```

see http://askubuntu.com/questions/446981/how-do-i-configure-my-upstart-scripts-to-take-dependencies-into-account

needs a fair bit of testing

enxebre:
#216 fix it

PR #186, opened 2015-05-14
Deploy docker/weave/registrator to all hosts

tayzlor:
Sets the weave network to be across the entire cluster and installs docker/registrator on the masters too. This should allow us to launch containers there (which we're probably going to want to do)

### DIFF
--- a/inventory/digitalocean/inventory
+++ b/inventory/digitalocean/inventory
@@ -17,5 +17,6 @@ mesos_masters
 mesos_masters
 
 [weave_servers:children]
+mesos_masters
 mesos_slaves
 load_balancers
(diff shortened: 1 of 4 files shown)


inventory/digitalocean/inventory
```diff
@@ -17,5 +17,6 @@ mesos_masters
 mesos_masters
 
 [weave_servers:children]
+mesos_masters
```
enxebre: If we'd use weave connect instead of launch, could we decouple the role form the inventory?

tayzlor: Could do. This could be fixed in #113

PR #187, opened 2015-05-14
only start slave mesos   when: mesos_install_mode == slave

enxebre:


tayzlor:
we should probably add a spec test to test that the slave isnt running on the master

Issue #182, opened 2015-05-13
Improve mesos spec tests


tayzlor:
We only test the service is running and the package is installed.

I'm fairly certain we could test for some of the files. Just need to be careful around the content of them as that is configurable, but we could test for the existence of them.

wallies:
could also test that there is a leader and there is correct number of masters of slaves

tayzlor:
This was partly fixed in https://github.com/Capgemini/Apollo/pull/191 (for the mesos master).

Just need the slave tests now

tayzlor:
Closing via #194

PR #189, opened 2015-05-14
increase quality of mesos master tests for issue #182

asnaedae:


tests/spec/mesos_master/mesos_master_spec.rb
```diff
@@ -4,10 +4,54 @@
...
+
+    describe file('/etc/mesos-master/quorum') do
+      it            { should be_file }
+      its(:content) { should contain '1' }
+    end
+  end
+
+  describe file('/etc/mesos/zk') do
+    it            { should be_file }
+    its(:content) { should contain('zk://172.31.1.1[123]:2181') }
```
wallies: will this work on environments other than vagrant

tayzlor: indeed this will break. For stuff thats dynamic (ip addresses) we should probably just test for the file, and maybe the content with some regex or something that is generic enough that its not hard-coded to a particular host so the test is portable to DO / AWS

asnaedae: Agreed, removed check for specific range, will add a more general regexp for IP (and now we have two problems)

enxebre: As Cam pointed don't we know this value ahead of time apart from vagrant

tests/spec/mesos_master/mesos_master_spec.rb
```diff
@@ -4,10 +4,53 @@
...
+  describe file('/etc/mesos-master') do
+    it { should be_directory }
+
+    describe file('/etc/mesos-master/ip') do
+      it { should be_file }
+    end
+
+    describe file('/etc/mesos-master/quorum') do
+      it            { should be_file }
+      its(:content) { should contain '1' }
```
tayzlor: This should normally be 2 (in a 3-node cluster)

tayzlor:
looks nice!

tests/spec/mesos_master/mesos_master_spec.rb
asnaedae: fixed.

Issue #181, opened 2015-05-13
Improve zookeeper spec tests


tayzlor:
we can add tests for 

- zookeeper conf file
- zookeeper id file
- zookeeper consul service description

tayzlor:
Closing via #192

PR #192, opened 2015-05-14
#181 - zookeeper spec tests

tayzlor:


tests/spec/zookeeper/zookeeper_spec.rb
```diff
@@ -7,3 +7,22 @@
 describe port(2181) do
   it { should be_listening }
 end
```
asnaedae: How about a command() test for "echo ruok | nc 127.0.0.1 2181"?

tayzlor: yep, will add it in, cheers!

PR #194, opened 2015-05-14
add mesos client tests #182

asnaedae:


tests/spec/mesos_slave/mesos_slave_spec.rb
```diff
@@ -4,6 +4,30 @@
   it { should be_installed }
```
tayzlor: should we add a test here to ensure the master service is not running on the slave?

asnaedae: good plan.

tests/spec/mesos_slave/mesos_slave_spec.rb
```diff
@@ -4,6 +4,30 @@
...
+    it            { should be_file }
+    its(:content) { should match /^MASTER=`cat \/etc\/mesos\/zk`$/ }
+  end
+
+  describe file('/etc/mesos-slave') do
+    it { should be_directory }
+  end
+  describe file('/etc/mesos-slave/work_dir') do
+    it            { should be_file }
+    its(:content) { should contain '/tmp/mesos' }
```
wallies: i dont know whether this will always be /tmp/mesos, on the other hand we should just test for defaults

tayzlor: yeah will break if someone configures it differently but should be ok within the confines of a test harness

asnaedae: will strip it out for now, doesn't really add a huge amount of confidence by being there.

Issue #176, opened 2015-05-13
Improve consul spec tests


tayzlor:
We can add the following checks - 

- consul ports - DNS/Leader/HTTP
- consul config files
- override files have gone

tayzlor:
Closed via https://github.com/Capgemini/Apollo/pull/195

PR #195, opened 2015-05-14
Made a start to #176 - consul specs

tayzlor:
Made a start on this. 

We need to think about how we are going to split tests for agent/server mode (which have slightly different configs). Any thoughts?


asnaedae:
Benefit of splitting the tests would be using host() and be_reachable which can check that there's connectivity between nodes (or correct firewall rules)

tayzlor:
LGTM

Issue #109, opened 2015-04-27
Create an AWS public cloud terraform plan


tayzlor:
We already have the VPC. lets do a pure public cloud version (similar to digitalocean one) for demo purposes.

tayzlor:
Closing via https://github.com/Capgemini/Apollo/pull/198

PR #198, opened 2015-05-15
Initial commit #109 - aws public TF plan

tayzlor:
Still work to do here, opening for initial feedback and if anyone else feels like completing it.
Guess this will just be used primarily for demos / test harness. Maybe we could even lower the barrier here and do away with Atlas integration?

- [ ] Test TF plan (and fix any outstanding bugs)
- [x] Add bootstrap scripts to provision on AWS
- [ ] Add docs


tayzlor:
tagging tests as might be handy for an integration test with AWS.

wallies:
This might cause you an issue
security_groups - (Optional) A list of security group IDs or names to associate with. If you are within a VPC, you'll need to use the security group ID. Otherwise, for EC2, use the security group name.

wallies:
got it working. Only bug is when you run bootstrap/apollo-down.sh it asks for values, is this expected?

enxebre:
"Only bug is when you run bootstrap/apollo-down.sh it asks for values" is unrelated see https://github.com/Capgemini/Apollo/pull/219/files

enxebre:
See https://github.com/Capgemini/Apollo/pull/234

tayzlor:
I've got the following issues - 

**AWS VPC**

- No nodes/services available in Consul UI. 

**Digitalocean** 

- No slaves Registered in Mesos
- Mesos redirect to master redirects to the hostname instead of the IP address

PR #202, opened 2015-05-17
Add note in vagrant docs about sudo

wallies:


tayzlor:
can we not bake this into the scripts?

I'm also wondering a couple of things... Should we -

- Lock on minimum vagrant version https://docs.vagrantup.com/v2/vagrantfile/vagrant_version.html (i think most of us have been testing on 1.7.2?)
- Add an if around ```override.vm.provision :hosts``` to test if the vagrant-hosts plugin is installed
- Forward the ports for the web interfaces in the Vagrantfile (incase the vagrant-hosts plugin is disabled?)

wallies:
Was going to bake it into the scripts, but wasnt sure. I'll add a minimum vagrant version of 1.7, as thats when things changed in the config.ssh. Will also add the hosts test as well

tayzlor:
lets do min version and the if wrapper for now then and just a docs fix

wallies:
Add check for vagrant and if wrapper. Didnt do port forwarding as we have the hosts on private network

docs/getting-started-guides/vagrant.md
```diff
@@ -24,6 +24,8 @@ export APOLLO_PROVIDER=vagrant
 sh bootstrap/apollo-launch.sh
 ```
 
+Note: You will either have to run the above command as sudo or have your hosts file writeable for the user running the command
```
tayzlor: does this doc only apply if using the vagrant-hosts plugin? might want to put a note in there about that if so.

tayzlor:
Code here LGTM, docs look a bit suspect.

asnaedae:
Aren't we already checking for if the plugin is installed, so isn't the check for the plugin during the rest of the vagrant file redundant?

  # throw error if vagrant-hosts not installed
  unless Vagrant.has_plugin?("vagrant-hosts")
    raise "vagrant-hosts plugin not installed"
  end

PR #203, opened 2015-05-17
Capture only first row of terraform --version output

enxebre:


enxebre:
Terraform version returns "Terraform v0.5.0

Your version of Terraform is out of date! The latest version
is 0.5.1. You can update by downloading from www.terraform.io"

When not using latest version. Need to suit script for capturing v0.5.0

PR #206, opened 2015-05-18
pipelining=True timeout = 15 ins ansible.cfg for performance improvement.

enxebre:


tayzlor:
Do we want to tweak - http://docs.ansible.com/intro_configuration.html#forks ?

Also should we enable ```gather_hosts: smart``` http://docs.ansible.com/intro_configuration.html#gathering ?

Your formatting is a bit weird on the pipelining bit (no spaces either side of the =)

enxebre:
Increase forks could potentially cause more network issues as more network network will be required by default. The default value is 5 and we only have 3 hosts (by default) that we want to run in parallel so 5 should be enough.

tayzlor:
:+1:

Issue #180, opened 2015-05-13
Improve weave spec tests


tayzlor:
At the moment we only test the bridge. We could -

- test for the container
- test docker options for weave
- test the network interfaces file

tayzlor:
Closing via https://github.com/Capgemini/Apollo/pull/208

PR #208, opened 2015-05-18
#180 - Add weave spec tests

tayzlor:


roles/weave/tasks/main.yml
```diff
@@ -27,14 +30,18 @@
...
     - weave
 
 - name: configure weave bridge for docker
   sudo: yes
-  lineinfile: "dest=/etc/default/docker state=present regexp=^DOCKER_OPTS= line='DOCKER_OPTS=\"{{ weave_docker_opts }}\"'"
+  lineinfile:
+   dest: /etc/default/docker
+   state: present
+   regexp: ^DOCKER_OPTS=
+   line: DOCKER_OPTS=\"{{ weave_docker_opts }}\"
```
wallies: Will this still work if we add docker options in other areas e.g. private dockerhub, storage options etc.

tayzlor: potentially not but we can cross that bridge when we come to it

PR #216, opened 2015-05-19
moving weave into and upstart script

enxebre:


roles/weave/templates/weave.conf.j2
```diff
@@ -0,0 +1,21 @@
...
+ [ -e /etc/default/weave ] && . /etc/default/weave
+ ${WEAVE} launch ${PEERS}
+ exec /usr/bin/docker logs -f weave
+end script
+
+pre-stop exec ${WEAVE} stop
+
+respawn
+respawn limit 10 10
+kill timeout 10
```
tayzlor: needs a new line EOF

roles/weave/tasks/main.yml
```diff
@@ -6,19 +6,13 @@
...
     - weave
 
-- name: Start up docker
-  service:
-    name: docker
-    state: started
-  tags:
-    - weave
+# Start docker as it is a requirement for weave create-bridge.
+- service: name=docker state=started
```
tayzlor: can we keep this like - 

```
service:
  name: docker
  state: started
```

roles/weave/tasks/main.yml
```diff
@@ -29,21 +23,29 @@
   tags:
     - weave
 
+# Create weave bridge.
 - name: bring up weave bridge
   command: ifup weave
   sudo: yes
+
+- name: upload weave template service
```
tayzlor: this gets uploaded here (will the service get started, as part of some notify handler?) might be better to jig the order so this is before the docker start (incase there is no notify handler)??

enxebre: The script is uploaded here. Dont need any notify as it has:
"start on started docker"
It is only going to be run onces docker restart in the next task

tayzlor:
LGTM

roles/weave/tasks/main.yml
```diff
@@ -6,19 +6,15 @@
...
   tags:
     - weave
 
-- name: Start up docker
-  service:
+# Start docker as it is a requirement for weave create-bridge.
+- service:
     name: docker
     state: started
-  tags:
```
tayzlor: any reason to remove the tags here?

tayzlor:
just realised we probably need to update the spec tests here

PR #221, opened 2015-05-20
Udacity bootstrap aws fixes

pmbauer:
1) `-F ssh.config` needs set in the ProxyCommand too, otherwise we can't
use the nat as an ssh proxy, 10.* addresses get added to known_hosts on
each run, poluting that file for subsequent runs, and the ProxyCommand
attempts to negotiate host key checking.

Environment:
Ubuntu 15.05
OpenSSH_6.7p1 Ubuntu-5ubuntu1, OpenSSL 1.0.1f 6 Jan 2014

2) make sed command portable


tayzlor:
@pmbauer thanks! This might also help you https://github.com/Capgemini/Apollo/commit/eeeb6089f644a4b72265fe838a8f6a30d0ab8640 (in latest master)

pmbauer:
Yep, thanks. Using eeeb608

PR #224, opened 2015-05-22
[roles/registrator] remove vestigial hostname

pmbauer:
- setting `hostname` is incompatible with `net: host`
- `ansible_all_ipv4_addresses` yields an invalid value for `hostname`


tayzlor:
Awesome, thanks!

PR #226, opened 2015-05-22
[roles] use service module for starts/restarts

pmbauer:
The shell module as used is not idempotent, use the service module -
built-in idempotency and portable.


tayzlor:
Awesome, thanks!

PR #227, opened 2015-05-22
[roles/docker+weave] opt. set graph and tmp dirs

pmbauer:
- this is optional and defaults are only  over-ridden if the following
  env vars are defined:
        - `APOLLO_docker_graph_dir`
        - `APOLLO_docker_tmp_dir`
- needed because for many ec2 instances, the large ephemeral volume is
  mounted under /mnt and the boot volume is very small
- rework the lineinfile calls to be idempotent
- rework the lineinfile call in docker and weave so
  DOCKER_OPTS directive is accumulative

### DIFF
--- a/roles/docker/defaults/main.yml
+++ b/roles/docker/defaults/main.yml
@@ -1,2 +1,4 @@
 ---
 # defaults file for docker
+docker_graph_dir: /var/lib/docker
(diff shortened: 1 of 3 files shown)


tayzlor:
image

PR #230, opened 2015-05-22
Automated deployment and specs from wercker

enxebre:
This is so we can discuss automated deploy/specs aproach.
Need to set correct artifact name and version.


enxebre:
Created follow up for set up aws https://github.com/Capgemini/Apollo/issues/231
Created follow up for moving account https://github.com/Capgemini/Apollo/issues/232

Issue #66, opened 2015-04-10
Have mesos span multiple availability zones


tayzlor:
![screen shot 2015-04-10 at 09 46 37](https://cloud.githubusercontent.com/assets/1681896/7084576/a7000cc0-df66-11e4-85df-b746410f4a17.png)

We should be able to achieve something like the above - having 2 masters in one AZ and another in a second AZ (providing higher availability)

keithchambers:
To improve availability you'd need to spread ZK and Consul across 3 AZs in a single region.

tayzlor:
![mesos-ha-architecture](https://cloud.githubusercontent.com/assets/1681896/8307068/7f1c6388-19b4-11e5-900f-cc99de1f779e.png)

keithchambers:
This should work fine.

enxebre:
Closing as https://github.com/Capgemini/Apollo/pull/385 got in.

PR #241, opened 2015-05-25
multi az support

wallies:


wallies:
this should solve #66

roles/consul/defaults/main.yml
```diff
@@ -5,7 +5,8 @@ consul_dc: dc1
 consul_servers_group: consul_servers
 consul_advertise: "{{ ansible_default_ipv4.address }}"
 consul_bind_addr: "{{ ansible_default_ipv4.address }}"
-consul_bootstrap_expect: "{{ groups[consul_servers_group] | length }}"
+consul_retry_join: "{% for host in groups[consul_servers_group] | intersect(groups[consul_dc]) %}\"{{ hostvars[host].ansible_default_ipv4.address }}\"{% if not loop.last %}, {% endif %}{% endfor %}"
```
tayzlor: whats the significance of adding the retry join?

terraform/aws/elb.tf
```diff
@@ -0,0 +1,24 @@
...
+    healthy_threshold = 2
+    unhealthy_threshold = 2
+    timeout = 3
+    target = "HTTP:80/"
+    interval = 30
+  }
+
+  instances = ["${aws_instance.mesos-slave.*.id}"]
+  cross_zone_load_balancing = true
+}
```
tayzlor: newline EOF needed

terraform/aws/variables.tf
```diff
@@ -43,9 +43,20 @@ variable "public_subnet_cidr_block" {
   default     = "10.0.0.0/24"
 }
 
-variable "private_subnet_cidr_block" {
-  description = "Cidr block for private Mesos subnet."
-  default = "10.0.1.0/24"
+variable "zones" {
+    default = {
+        zone0 = "eu-west-1a"
```
tayzlor: for styling / consistency might be worth either keeping all of these like ```zone0``` or moving to ```zone-0``` like the master_ips var. Otherwise move the master_ips to be like ```master1```

tayzlor:
Initial overview of the code looks like it is good. Can we also update the architecture diagram on the frontpage with this change? 

To do that you need to provision the cluster as normal, then login to visualops then there is an option to import from AWS (and you can import the running instances) which should create a diagram which you can view/edit

roles/consul/defaults/main.yml
wallies: consul was failing to elect a leader and then failing to show up as services in the UI. Adding this fixed it.
-retry-join - Similar to -join but allows retrying a join if the first attempt fails. This is useful for cases where we know the address will become available eventually.

pmbauer: fwiw, I've seen consul consistently fail to cluster on `bootstrap/apollo-launch.sh` and every consul agent ends up with a cluster of one member, itself.  We worked around it by defining `consul_join` but this is more flexible.

tayzlor: @pmbauer I've seen that too occassionally - I reported it over in https://github.com/Capgemini/Apollo/pull/198. Looks like the consul related fixes here might address that too

tayzlor: can we separate the consul retry join to a separate PR as that may resolve issues over in #198 and to avoid blocking on waiting for multi-az

wallies:
couldnt add variables, as cant have dots in bash variables
export TF_VAR_zones.zone-0=${TF_VAR_zones.zone-0:-eu-west-1a}
export TF_VAR_zones.zone-1=${TF_VAR_zones.zone-1:-eu-west-1b}
export TF_VAR_zones.zone-2=${TF_VAR_zones.zone-2:-eu-west-1c}

enxebre:
Closing in favour of https://github.com/Capgemini/Apollo/pull/385

PR #242, opened 2015-05-25
239 ssh

enxebre:
This is for checking ssh availabilty before gathering facts and also after from an ansible task.
It's been tested succesfully however the behaviour is different from wercker and the deploy script just skip the next step after running the check ssh availabilty script. See https://app.wercker.com/#deploystep/5563a23998eb5da126072b45

### DIFF
--- a/site.yml
+++ b/site.yml
@@ -1,5 +1,6 @@
 ---
 - hosts: all
+  gather_facts: no
   tasks:
     - name: Wait for port 22 to become available.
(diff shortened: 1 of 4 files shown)


tayzlor:
Is it necessary to have both the bash implementation and the ansible implementation here?

site.yml
```diff
@@ -1,5 +1,9 @@
 ---
 - hosts: all
```
tayzlor: should this block have - 

```
gather_facts: false
```

PR #257, opened 2015-05-26
Consul fix

wallies:
Retry join is more effective when joining and electing leader


Vagrantfile
```diff
@@ -79,6 +80,8 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
             mesos_local_address: node[:ip],
             marathon_local_address: node[:ip],
             consul_join: consul_join,
+            consul_retry_join: [consul_retry_join],
+            consul_bootstrap_expect: 1,
```
tayzlor: should this not be 3? we're still provisioning 3 masters in the vagrant setup

tayzlor: if it should be 3, no need to specify it as it will be 3 servers in the group (above in ansible groups)

wallies: if i specify 3, then it fails waiting for leader on master1

Issue #218, opened 2015-05-19
Replace ansible dynamic inventories per cloud with 1 inventory that looks at the terraform state file


tayzlor:
It's a bit cumbersome to have to rinse and repeat per-cloud provider. We should be able to craft a solution thats a single entry-point for all the clouds via the tfstate file

enxebre:
This needs to be smart enough for grouping by dc see https://github.com/Capgemini/Apollo/pull/259

PR #259, opened 2015-05-27
Creating dc1 group in inventory

enxebre:
We need to move the population of this group to the dynamic inventory.
See https://github.com/Capgemini/Apollo/issues/218


wallies:
going to have to change consul.json.j2 from consul_dc to consul_dc_group

Issue #255, opened 2015-05-26
Upgrade to weave 0.11.1


tayzlor:
https://github.com/weaveworks/weave/releases/tag/v0.11.1

tayzlor:
Closing, superseded by https://github.com/Capgemini/Apollo/issues/317

PR #210, opened 2015-05-19
Bump to consul 0.5.2

tayzlor:

PR #262, opened 2015-05-27
#255 - upgrade weave to v0.11.2

tayzlor:


wallies:
if you are going to rebuild the packer builds, remember there is the consul upgrade as well

tayzlor:
Should go with #210 . Tag a new version / release new packer builds.

tayzlor:
This is out of date as 0.11.1 has since been released

PR #264, opened 2015-05-28
Use vagrant-hostmanager instead of vagrant-hosts plugin

tayzlor:
This plugin actually supports updating the host machines /etc/hosts (the other one only does the guests). This plugin also more supported/used in the community and seems to work better than vagrant-hosts


wallies:
the one we have was updating my host file, it used the unique id of the box

Issue #229, opened 2015-05-22
Add weave scope to be provisioned with the weave role


tayzlor:
We should be able to provision - https://github.com/weaveworks/scope along with weave

PR #265, opened 2015-05-28
Issue #229 Add support for weave scope

tayzlor:


tayzlor:
Adding a note here to launch with peers

Issue #275, opened 2015-06-02
ELB not balancing correctly to slave node haproxy containers


tayzlor:
![screen shot 2015-06-02 at 09 22 48](https://cloud.githubusercontent.com/assets/1681896/7931489/00260978-0909-11e5-8db5-f03de186a8d2.png)

This is because it tries to do a health check to the slave IP address on port 80 which returns a 503 (because there is no backend set up in haproxy to return from that)

tayzlor:
We need to clarify how we're using the ELB here. We might want to do something like this https://github.com/jvehent/haproxy-aws

tayzlor:
This Pull request adds support for the Proxy protocol in the ELB https://github.com/hashicorp/terraform/pull/1749/

enxebre:
closing this as per https://github.com/Capgemini/Apollo/commit/785e790d1c623f73d33029dd21b74779873b2056

PR #281, opened 2015-06-02
Initial commit on #275 - Fix the ELB

tayzlor:
- Fix health check for ELB
- Expose health check port on haproxy container
- Use PROXY mode on ELB in AWS

This fix needs more testing

### DIFF
--- a/roles/haproxy/files/haproxy.tmpl
+++ b/roles/haproxy/files/haproxy.tmpl
@@ -19,7 +19,7 @@ frontend haproxy_status
     monitor fail if backend_dead
 
 frontend www
-    bind *:80 accept-proxy
+    bind *:80
 
     # Generated automatically by consul-template
 {{ range services }}
--- a/roles/haproxy/files/haproxy.tmpl
+++ b/roles/haproxy/files/haproxy.tmpl
@@ -1,15 +1,25 @@
 global
-    maxconn 256
+    maxconn {{or (key "service/haproxy/maxconn") 256}}
     debug
 
 defaults
     mode http
-    timeout connect 5000ms
-    timeout client 50000ms
-    timeout server 50000ms
+    timeout connect {{or (key "service/haproxy/timeouts/connect") "5000ms"}}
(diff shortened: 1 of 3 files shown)


roles/haproxy/files/haproxy.tmpl
```diff
@@ -1,15 +1,25 @@
...
+    # create a status URI in /haproxy_status that will return
+    # a 200 is backend is healthy, and 503 if it isn't. This
+    # URI is queried by the ELB.
+    acl backend_dead nbsrv(consul_backend) lt 1
+    monitor-uri /haproxy_status
+    monitor fail if backend_dead
 
 frontend www
-    bind *:80
+    bind *:80 accept-proxy
```
tayzlor: There appears to be a problem with this line here (when running up in the vagrant machine/aws, so removing it for now)

wallies:
Does this need terraform 0.5.3, if so you should add that to the terraform version check

tayzlor:
I ran it up on 0.5.2

wallies:
Does it require 0.5.2 then, if so, can we add that check to the terraform version

tayzlor:
I can't seem to test this fix properly. Here's what I tried, pasting here incase anyone else wants to have a go - 

- provisioned the instances
- got the ELB hostname and did - ```dig +short apollo-elb-578724806.eu-west-1.elb.amazonaws.com``` as exmaple. You should get the IP(s) of the ELB
- Tried running a ```curl -s -v -H 'Host: marathon.example.com' http://54.76.83.152```

My end goal was to get an IP and stick it in ```/etc/hosts``` and access that statically mapped host through my browser.

pmbauer:
I discovered what the problem is.
https://github.com/Capgemini/Apollo/commit/33e53fb56a1c1486dad761428ba481ce4c05332e#diff-ee33aa395b84461181b557096d97ed52R8

The listener here needs to be 'http' protocol.
Then I was able to successfully run `curl -v -H 'Host: marathon.example.com' http://apollo-mesos-elb-XXXXXXX.us-west-1.elb.amazonaws.com`

I kept the health check listener the same as in 33e53fb.

PR #283, opened 2015-06-02
Add proxy support to vagrant setup

wallies:


group_vars/vagrant
```diff
@@ -16,3 +16,10 @@ weave_launch_peers: "
   {%- endfor -%}
   {{ weave_peers|join(' ') }}
 "
+proxy_env: {
```
enxebre: Does it need "{ }" ?

Issue #280, opened 2015-06-02
Consul retry_join not formatted correctly


tayzlor:
According to the consul docs - https://www.consul.io/docs/agent/options.html#_retry_join

it says 
> Takes a list of addresses to attempt joining 

If you provision the vagrant box you get a single array element with a string separated by commas - 

```
  "retry_join": [ "172.31.1.11, 172.31.1.12, 172.31.1.13" ],
```

I think this should be - 

```
 "retry_join": [ "172.31.1.11", "172.31.1.12", "172.31.1.13" ],
```

so 3 array elements instead of 1

tayzlor:
Closing via #284

PR #284, opened 2015-06-03
Fix dependencies

### LINKED ISSUE #278: Set consul_is_server variable dynamically
At the moment the default in ```consul/defaults/main.yml``` is 

```
consul_is_server: no
```

We should be able to set this to yes/no programmatically by checking if the current host is in the ```consul_servers_group```

At the very least if we can't do that (don't see why we shouldn't be able to) we can create - 

```
groups_vars/consul_servers
```

and set ```consul_is_server: yes``` in there

Then we can remove the setting from the site.yml

tayzlor:
This fixes #278 and also should speed up provisioining because we don't constantly re-run some roles on masters/slaves.


tayzlor:
Also added a fix for #280

PR #287, opened 2015-06-03
GCE support.

enxebre:

PR #290, opened 2015-06-03
Bash refactor.

enxebre:
This needs https://github.com/Capgemini/Apollo/pull/287 to be merged as this remove cloud credentials from ansible run so need terraform inventory.
I got Unable to locate package shellcheck when trying to install it from wercker.


tayzlor:
Regarding shellcheck you'd need to check the universe repo is enabled http://installion.co.uk/ubuntu/utopic/universe/s/shellcheck/install/index.html

Don't know if that is possible through wercker?

bootstrap/aws-public/config-default.sh
```diff
@@ -13,6 +13,8 @@ export TF_VAR_key_name=${TF_VAR_key_name:-apollo}
 # so no really needed to export them. Exporting for consitency.
 export TF_VAR_atlas_artifact_master=${TF_VAR_atlas_artifact_master:-capgemini/apollo-ubuntu-14.04-amd64}
 export TF_VAR_atlas_artifact_slave=${TF_VAR_atlas_artifact_slave:-capgemini/apollo-ubuntu-14.04-amd64}
+export TF_VAR_atlas_artifact_version_master=${TF_VAR_atlas_artifact_version_master:-1}
```
tayzlor: is this necessary? seems to work fine without it at the moment

bootstrap/aws/util.sh
```diff
@@ -44,7 +20,7 @@ apollo_launch() {
 }
 
 ansible_ssh_config() {
-  pushd $APOLLO_ROOT/terraform/aws
+  pushd "${APOLLO_ROOT}/terraform/${APOLLO_PROVIDER}"
     BASTION_IP=$(terraform output bastion.ip)
```
tayzlor: we ```export APOLLO_bastion_ip``` earlier, then we get from terraform again, could we just use APOLLO_bastion_ip?

bootstrap/vagrant/config-default.sh
```diff
@@ -1 +1,2 @@
 #!/bin/bash
+echo ""
```
tayzlor: ?

enxebre: commited by accident, shellchecker was complaining about empty file, probably want to remove that file and check if exits before source it

bootstrap/aws/util.sh
enxebre: yep we could do export  APOLLO_bastion_ip=$(terraform output bastion.ip)

bootstrap/aws-public/config-default.sh
enxebre: As we'll be using a generic terraform apply for all the clouds, yes it is necessary. Also it is working bacuse we are relying on variables.tf in some cases and conofig-default.sh for others, we should probably have a better convention/consitency. Also we could refactor config-default files in a follow up pr as we have a lot of duplication.

enxebre:
Thanks for shellchecker link, how ever seems like universe is already there:
$ sudo cat /etc/apt/sources.list
deb http://archive.ubuntu.com/ubuntu precise main restricted universe multiverse

PR #300, opened 2015-06-11
AWS public wercker multiple deploy.

enxebre:


tayzlor:
I think we need to bump the terraform version for wercker

PR #309, opened 2015-06-15
HVM Virtualisation for AMI

wallies:


tayzlor:
Is the 

``` 
"source_ami": "{{user `aws_source_ami`}}",
```

in affect here as well? Do you have to build on top of a HVM source AMI ?

wallies:
Yep source ami stays the same

PR #311, opened 2015-06-17
fix awk error in get_apollo_variables

pmbauer:
This prints a literal `${1}` instead of the value of the first column.

bash refactor commit broke passing APOLLO_* env vars to ansible
5cf39b6a5366d3084c4e113926178f3c7ed34a00
https://github.com/Capgemini/Apollo/blob/5cf39b6a5366d3084c4e113926178f3c7ed34a00/bootstrap/common.sh#L62


enxebre:
nice catch, thanks again!

Issue #315, opened 2015-06-19
Race condition with Docker (devicemapper) and Udev sync


tayzlor:
On provisioning we occassionally hit this error - 

```
msg: Docker API Error: Cannot start container db72748b5f7e8bc004f7d7f99164098b055f6d8bbf4a3c2a2945d28237ad6884: Error getting container db72748b5f7e8bc004f7d7f99164098b055f6d8bbf4a3c2a2945d28237ad6884 from driver devicemapper: Error mounting '/dev/mapper/docker-202:1-274961-db72748b5f7e8bc004f7d7f99164098b055f6d8bbf4a3c2a2945d28237ad6884' on '/var/lib/docker/devicemapper/mnt/db72748b5f7e8bc004f7d7f99164098b055f6d8bbf4a3c2a2945d28237ad6884': no such file or directory
```

Running some tests on a machine in AWS - 

```
docker info | grep Udev
 Udev Sync Supported: false
WARNING: No swap limit support
root@ip-10-0-0-225:~# ldd /usr/bin/docker
	not a dynamic executable
```

This is a bug that is detailed here - https://github.com/docker/docker/issues/4036

The workaround fix is described here - http://icarobichir.com.br/posts/building-a-dynamically-linked-docker-v-1-6-2/

Since docker 1.7.0 is out we should probably upgrade and see if that helps too (see #316)

pmbauer:
See also: https://github.com/docker/docker/issues/12535

This error is mitigated when using aufs instead of devicemapper.
At Udacity, we are building our own image, forcing aufs instead of devicemapper.
https://github.com/udacity/Apollo/commit/a9825c43b24493c414a6ee05cd45150823dfbfee

This issue only affects ubuntu 14.04 (server).  Desktop Ubuntu has the aufs drivers out-of-the-box, so if you install docker on an Ubuntu dev machine, it will default to aufs.  We wanted to normalize the storage driver in dev and prod environments and chose aufs.

_Note_: There are also performance implications for operating aufs vs devicemapper that caused us to lean towards aufs.
http://jpetazzo.github.io/assets/2015-03-03-not-so-deep-dive-into-docker-storage-drivers.html#36
http://jpetazzo.github.io/assets/2015-03-03-not-so-deep-dive-into-docker-storage-drivers.html#44.

pmbauer:
[...]

PR #319, opened 2015-06-19
[packer] make sure docker is using aufs

pmbauer:
(context https://github.com/Capgemini/Apollo/issues/315)
- aufs drivers aren't installed by default on Trusty cloud images, which
  causes docker to fall back to devicemapper.  This is bad because
  devicemapper is completely broken on Trusty


wallies:
Cheers @pmbauer

PR #322, opened 2015-06-21
More simple and dynamic vagrant configuration

lordoffreaks:


tayzlor:
Looks good on the face of things, will see if anyone else has anything to pitch in. WIll try give it a spin too when i've some spare time

lordoffreaks:
It looks like the bits
```
machine.vm.hostname = node[:hostname]
 machine.vm.network :private_network, :ip => node[:ip]
 
 vb.name = 'vagrant-mesos-' + node[:hostname]
 vb.customize ["modifyvm", :id, "--memory", node[:mem], "--cpus", node[:cpus] ]
```
Which are common to both masters and slaves could be unified to avoid repetiton but my guess is that's the scope of another PR

enxebre:
More improvements can be done in follow ups PR. LGTM

PR #327, opened 2015-06-22
Set default weave iprange parameter.

enxebre:


roles/weave/templates/weave.conf.j2
```diff
@@ -10,7 +10,7 @@ pre-start exec ${WEAVE} stop
 
 script
  [ -e /etc/default/weave ] && . /etc/default/weave
- ${WEAVE} launch ${PEERS}
+ ${WEAVE} launch "{{ weave_docker_subnet }}" ${PEERS}
```
tayzlor: does this not need to be - 

```
 ${WEAVE} launch -iprange "{{ weave_docker_subnet }}" ${PEERS}
```

PR #330, opened 2015-06-23
Reverting aws key name.

enxebre:


enxebre:
Terraform understand that there are changes and rebuild infrastructure.
Also there are more clashes during deployment. Need to create a better general solution for supporting parallel deployments  in to the same account

Issue #292, opened 2015-06-05
Upgrade to consul_template 0.9.0


wallies:
in roles/haproxy/main.yml

PR #323, opened 2015-06-22
upgrade docker, weave, marathon

wallies:

PR #333, opened 2015-06-23
#292 - Upgrade to consul_template 0.9.0

ravbaba:
Upgrading to consul_template 0.9.0 - issue #292


tayzlor:
0.10.0 is out so we should go to that, See https://github.com/hashicorp/consul-template/releases

wallies:
upgraded in #323

PR #334, opened 2015-06-23
Docker bench

wallies:


tests/test.yml
```diff
@@ -3,13 +3,16 @@
   gather_facts: no
   roles:
     - { role: ../roles/serverspec, test_role: "mesos_masters" }
```
enxebre: Do we need test_role variable for dockerbench task?

enxebre:
Need to confirm that dockerbench is going to return an exit code different than zero when detecting warnings/errors so it would make ansible fail so we can get the feedback from the wercker deploy

roles/dockerbench/tasks/main.yml
```diff
@@ -0,0 +1,20 @@
+---
```
enxebre: Can we add "when: dockerbench_run_test" here for this tasks

PR #338, opened 2015-06-24
Run marathon as a docker container.

enxebre:


roles/marathon/defaults/main.yml
```diff
@@ -1,8 +1,30 @@
 ---
 # defaults file for marathon
-marathon_port: 8080
-consul_dir: /etc/consul.d
+marathon_enabled: true
+marathon_version: 'v0.8.2'
```
tayzlor: I think we should use https://github.com/mesosphere/marathon/releases/tag/v0.9.0-RC2

roles/marathon/tasks/main.yml
```diff
@@ -1,104 +1,81 @@
...
-    src: marathon.conf.j2
-    dest: /etc/init/marathon.conf
-    mode: 0755
-  notify:
-    - restart marathon
-  sudo: yes
-  tags:
-    - marathon
+# Install (docker-py) python package as is a docker module dependency.
+- pip: name=docker-py version=1.1.0
```
tayzlor: dont think this is needed now its installed in packer?

tayzlor:
This looks pretty good. after the comments and a sqaush it should be good to merge.

tayzlor:
should we be removing the packer install scripts for the package in this PR too?

tayzlor:
Lets also tag on master before this gets in

roles/marathon/defaults/main.yml
enxebre: Would need to rebuild packer images if we dont want it to go the internet though

enxebre:
I will tag, then I would keep the packages in packer and merge this PR. I'll create a follow up for removing the packages and lastest marathon image version.
https://github.com/Capgemini/Apollo/issues/339

enxebre:
Also I noticed some perfomance issue when running marathon inside the container in vagrant that need to be investigated.

roles/marathon/tasks/main.yml
enxebre: Probablt dont neede but I like it as It helps to keep the role more self-contained

PR #342, opened 2015-06-25
Weave ip range.

enxebre:


lordoffreaks:
LGTM

tayzlor:
Hang fire on this as we're nt sure it's the right solution yet

pmbauer:
This may be related; this is just an anecdote and a small description of how we addressed the issue.

Due to assigning the same `weave_bridge` to every machine, we were seeing packet loss ~30% and extreme latency jitter.  This is because of gateway address collisions on each machine.

https://github.com/Capgemini/Apollo/blob/master/roles/weave/defaults/main.yml#L3

This is how we solved it:

https://github.com/udacity/Apollo/commit/e459832a30a097ba58cffb0b1c4710cc459b19be
Summary:
1) assign a unique weave_bridge (gateway) ip to each machine.  Masking ensures the same CIDER block
2) compute the weave subnet based on the VPC-assigned host IP.  This ensures elasticity - we don't need to re-allocate the weave subnets whenever slaves are added or removed.  This works because of carefully constructed VPC subnets and we effectively allow AWS to handle GC for us whenever they recycle a host IP address.

An alternative would be what coreos fabric is doing, namely store allocated subnet blocks in a consistent store.

tayzlor:
Closing this in favour of #351

PR #350, opened 2015-06-29
Make bashism conditional on the shell being bash

wallies:


enxebre:
There are probably more code making the our bash scripts not posix compliant so I think may be we should either create a ticket for making them posix compliant or just set bash as a requirement until we decide if we'll be using sh scripting for this in the future.

tayzlor:
agreed, would force them to run in bash

PR #351, opened 2015-06-30
Fix #342 - Weave network instability

tayzlor:
Addresses issues raised in #342.

Running some tests in aws-public result in much higher stability across the weave network

### DIFF
--- a/roles/weave/defaults/main.yml
+++ b/roles/weave/defaults/main.yml
@@ -1,6 +1,12 @@
 ---
 # defaults file for weave
-weave_bridge: "10.2.0.1/16"
+weave_bridge: "
+    {%- for host in groups[weave_server_group] -%}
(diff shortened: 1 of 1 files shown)


tayzlor:
Agreed this may not be the ideal though did fix a current issue where the weave network was pretty much unusable. 
We haven't given a whole heap of thought to elastic scaling yet from our POV at the moment. Have you any insights / thoughts about it? Maybe let's open a new issue and discuss it there.

pmbauer:
Great, wilco.
Weave was pretty much unusable previously, and I'm glad we weren't the only ones running into this issue.  We also solved it by computing different weave_bridge values for each host.

tayzlor:
Created https://github.com/Capgemini/Apollo/issues/355 to track this. Please pitch in with any thoughts over there so we can find a way forward

PR #365, opened 2015-07-03
fix chronos container hostname

pmbauer:
this fixes HA

chronos server registers itself with zookeeper using its hostname, which
is a random sha if not assigned.  As a result, redirects to the elected
master chronos server fail from both the gui and rest apis.


pmbauer:
closing for now; simply setting the hostname has the desired effect of having hot standby masters forward to the active masters but the unfortunate side effect of making chronos not work at all.  For now, the work-around is to determine the active chronos master and only communicate with that directly

PR #369, opened 2015-07-07
issue-258 - Create a single machine Vagrantfile

ravbaba:
Created a single machine Vagrantfile

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -4,6 +4,7 @@ require 'yaml'
 
 base_dir = File.expand_path(File.dirname(__FILE__))
 conf = YAML.load_file(File.join(base_dir, "vagrant.yml"))
+groups = YAML.load_file(File.join(base_dir, "ansible-groups.yml"))
 
 # Vagrantfile API/syntax version. Don't touch unless you know what you're doing!
 VAGRANTFILE_API_VERSION = "2"
@@ -29,11 +30,11 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   config.ssh.insert_key = false
 
   # Common ansible groups.
-  ansible_groups = conf['ansible_groups'];
+  ansible_groups = groups['ansible_groups']
+  ansible_groups["mesos_masters"] = []
 
   masters_conf = conf['masters']
-  masters_n = masters_conf['ips'].count
-  ansible_groups["mesos_masters"] = []
+  masters_n    = masters_conf['ips'].count
   master_infos = []
 
   # Mesos master nodes
@@ -41,11 +42,11 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
 
     ip = masters_conf['ips'][i - 1]
     node = {
-      :zookeeper_id    => i,
-      :hostname        => "master#{i}",
-      :ip              => ip,
-      :mem             => masters_conf['mem'],
-      :cpus            => masters_conf['cpus'],
+      :zookeeper_id => i,
+      :hostname     => "master#{i}",
+      :ip           => ip,
+      :mem          => masters_conf['mem'],
+      :cpus         => masters_conf['cpus'],
     }
 
     master_infos.push(node)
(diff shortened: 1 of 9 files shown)


group_vars/vagrant
```diff
@@ -5,8 +5,8 @@ consul_advertise: "{{ ansible_eth1.ipv4.address }}"
 consul_bind_addr: "{{ ansible_eth1.ipv4.address }}"
 marathon_hostname: "{{ ansible_eth1.ipv4.address }}"
 mesos_hostname: "{{ ansible_eth1.ipv4.address }}"
-mesos_master_quorum: 2
```
enxebre: This (mesos_master_quorum and consul_bootstrap_expect) could probably be removed from here now as we are  passing it explicitly.

Vagrantfile-standalone
```diff
@@ -0,0 +1,99 @@
...
+
+  node = {
+    :hostname => "standalone-cluster",
+    :ip => masters_conf['ip'],
+    :mem => masters_conf['mem'],
+    :cpus => masters_conf['cpus'],
+  }
+
+  master_infos.push(node)
+
```
enxebre: I think this would be useful if it would be dynamic but we are assuming as a prerequisite that is going to be only one host so dont think we need  to master_infos.push(node) and then map. We could just  use node or masters_conf, that would become host_conf or cluster_conf

Personally I'd be in favour of making a one dynamic Vagrantfile supporting any number of masters/slaves including the standalone. Then the yaml file with the settings could be set as an environment var. e.g vagrant_cluster=vagrant-standalone.yaml

Vagrantfile
```diff
@@ -116,7 +117,9 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
               consul_join: consul_join,
               consul_retry_join: consul_retry_join,
               marathon_master_peers: marathon_master_peers,
-              marathon_zk_peers: marathon_zk_peers
+              marathon_zk_peers: marathon_zk_peers,
+              mesos_master_quorum: 2,
```
lordoffreaks: Shouldn't these value be configurable through one of the `yaml` files?

enxebre:
We probably want to add how to use the standalone version in the Vagrant Docs.
Otherwise lgtm

I'd be in favour of making a one dynamic Vagrantfile supporting any number of masters/slaves including the standalone so we DRY. Then the yaml file with the settings could be set as an environment var. e.g vagrant_cluster=vagrant-standalone.yaml

Vagrantfile
ravbaba: do you mean those two? because rest of them are dynamic
mesos_master_quorum: 2
consul_bootstrap_expect: 3

But in my opinion those two should be in a vagrant.yml file

lordoffreaks: Yes ... those two ... I prefer them to live in `vagrant.yml` to avoid hardcoding values here

tayzlor:
The work here looks mostly fine, needs rebased and a couple of comments addressing. 
Some docs would be nice too + the ability to launch it via a bootstrap script (similar the multi-vm one would be good).

ravbaba:
I will try put the logic to one Vagrantfile as enxebre suggested and clean the code, I will update the documentation as well. Tayzlor I am gonna use the same script which we use now (apollo-launch.sh) everything what we need to do is to set an environment var which switch vagrant configuration between two files (vagrant.yml/vagrant-standalone.yml). 

Thank you for your answers.

Vagrantfile-standalone
```diff
@@ -0,0 +1,91 @@
...
+  # Common ansible groups.
+  ansible_groups = groups['ansible_groups'];
+  
+  ansible_groups["mesos_masters"] = []
+  ansible_groups["mesos_slaves"] = []
+  master_infos = []
+
+  node = conf['masters']
+
+  mesos_zk_url = "zk://"+node['ip']+":2181/mesos"
```
tayzlor: can you line all the = up so its a bit more readable here (same below)

bootstrap/common.sh
```diff
@@ -81,16 +81,24 @@ run_if_exist() {
...
-      /usr/bin/open "http://$(terraform output master.1.ip):5050"
-      /usr/bin/open "http://$(terraform output master.1.ip):8080"
-      /usr/bin/open "http://$(terraform output master.1.ip):8500"
-      /usr/bin/open "http://$(terraform output master.1.ip):4040"
-      /usr/bin/open "http://$(terraform output master.1.ip):4400"
-      /usr/bin/open "http://$(terraform output master.1.ip):8081"
-    fi
-  popd
+  local master_url=''
+
```
enxebre: could we move this to a separate function like get_master_url

tayzlor:
PR looks ready to merge, but looks like it needs rebased due to some conflict first. If you do that, we can merge it

ravbaba:
I've just fixed the conflict. Tayzlor, could you merge it.

PR #370, opened 2015-07-08
Vault support.

enxebre:
This provides:
Vault High availability set up.
Consul secret backend integration
App-id Vault auth backend allowing the users range to be set by env variables.
Vault ACL policy for weave.
Security weave setup up using Vault for retrieving a given password. 

Needs some thinking and discussion


roles/vault/tasks/auth.yml
```diff
@@ -0,0 +1,30 @@
...
+  run_once: true
+  when: vault_root_token is defined
+  environment:
+    VAULT_ADDR: "{{ vault_addr }}"
+  command: "vault auth-enable app-id"
+  sudo: yes
+  tags:
+    - vault
+
+- name: create weave app-id
```
tayzlor: should this be stored with the weave role ?

roles/vault/tasks/bootstrap.yml
```diff
@@ -0,0 +1,59 @@
...
+- name: unseal vault
+  when: vault_keys is defined
+  sudo: yes
+  command: "{{ vault_config_folder }}/unseal"
+  tags:
+    - vault
+
+- name: store vault security values
+  when: vault_root_token is defined and vault_initialized_output.stdout == "{\"initialized\":false}"
+  local_action: "template src=vault-security.yaml.j2 dest=./vault-security.yaml mode=0644"
```
tayzlor: lets format here on separate lines for readability, same in a couple of other places you've done inline formatting

roles/vault/tasks/install.yml
```diff
@@ -0,0 +1,74 @@
+- name: download vault binary
```
tayzlor: could these 2 commands be compressed to - 

```
# Unarchive a file that needs to be downloaded
- unarchive: src=https://example.com/example.zip dest=/usr/local/bin copy=no
```

roles/weave/templates/weave.conf.j2
```diff
@@ -10,7 +10,11 @@ pre-start exec ${WEAVE} stop
 
 script
  [ -e /etc/default/weave ] && . /etc/default/weave
- ${WEAVE} launch ${PEERS}
+
+ pass_json=$( curl -sL -H 'X-Vault-Token: {{ tmp_client_token }}' {{ vault_addr }}/v1/secret/weave/password )
```
tayzlor: do we want to have a hard dependency on vault here?

tayzlor:
in general this PR looks ok. my only question is should we move the weave stuff to live with the weave role?

enxebre:
Yes, We probably want to set up vault basis in its role and then every role will be responsible for handle its security tasks may be in a security.yml. How ever this need some more testing as after restarting consul and vault, sometimes they fall into an inconsistent state for a few seconds making other tasks fail.

tayzlor:
:+1:

enxebre:
Pending on https://github.com/hashicorp/vault/issues/311

tayzlor:
Removing the blocked tag for now we need to upgrade over to vault 0.3.1 and re-test this

tayzlor:
On the todo list for this i've got - 

- [ ] Reroll patch to fix conflicts
- [ ] Address comments / inline styling
- [ ] Upgrade to use vault 0.3.1
- [ ] Re-check the weave encryption approach with latest version of weave and update if necessary. Move the weave vault integration to sit with the weave role?

Anything else?

enxebre:
This is not feasible anymore as weave encryption is not compatible with fastdatapath.

tayzlor:
That's fine - Guess we just remove the weave encrypt stuff form this PR then. I think we can still get vault in anyway, as it should help with other security aspects

roles/vault/tasks/policies.yml
```diff
@@ -0,0 +1,21 @@
...
+  run_once: true
+  when: vault_root_token is defined
+  environment:
+    VAULT_ADDR: "{{ vault_addr }}"
+  command: "vault auth {{ vault_root_token }}"
+  sudo: yes
+  tags:
+    - vault
+
+- name: write acl weave policy
```
wallies: will this work?

roles/vault/tasks/secrets.yml
```diff
@@ -0,0 +1,11 @@
+- name: create weave password
```
wallies: not sure we need this anymore

docs/components/vault.md
```diff
@@ -0,0 +1,29 @@
+## Vault
+
+We use [Vault](https://vaultproject.io/) for for securely accessing secrets.
+
+The main decision behind choosing Vault is its easy integration with Consul as a [secret backend](https://vaultproject.io/docs/secrets/index.html).
+
+Vault provides "a unified interface to any secret, while providing tight access control and recording a detailed audit log".
+
+We use [App-id auth backend](https://vaultproject.io/docs/auth/app-id.html) as "a mechanism for dynamic machines, containers, etc. to authenticate with Vault. It works by requiring two hard-to-guess unique pieces of information: a unique app ID, and a unique user ID".
```
tayzlor: docs here need updated

tayzlor:
Do we want to provide a config switch around vault or we see it as a core piece of the platform ? (my opinion is of the latter but im just putting this out there)

wallies:
core part of the platform in my opinion

enxebre:
I'm good with that

tayzlor:
we reckon this is good to go? /cc @enxebre

wallies:
this looks good to me now. @enxebre are you fine with it?

enxebre:
Code looks good to me, have not had the chance to test it though.

tayzlor:
Changing to needs work - we need to bump to vault 0.4 and retest

roles/vault/tasks/install.yml
wallies: this isnt supported until ansible 2.0

tayzlor:
some build failures here in https://app.wercker.com/#buildstep/566efde6b5a450f41f00da7d

PR #371, opened 2015-07-09
Added chronos_zk_url variable to ansible.extra_vars

ravbaba:
There was an issue with running chronons on VM. It was because of we didn't pass chronos_zk_url into ansible variables.


enxebre:
thanks!

PR #378, opened 2015-07-12
Providing dynamic ip ranges different than 10.* for AWS based on vpc_cidr_block

lordoffreaks:
This PR removes the `10.` hardcoded value for AWS to provide a dynamic behaviour based on the value of the `aws_vpc.default.cidr_block` variable

### DIFF
--- a/bootstrap/aws/util.sh
+++ b/bootstrap/aws/util.sh
@@ -17,6 +17,10 @@ ansible_ssh_config() {
   pushd "${APOLLO_ROOT}/terraform/${APOLLO_PROVIDER}"
     export APOLLO_bastion_ip=$(terraform output bastion.ip)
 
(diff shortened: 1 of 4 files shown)


tayzlor:
looks ok to me will wait for another review and merge

bootstrap/aws/util.sh
```diff
@@ -84,3 +85,7 @@ ovpn_client_config() {
     done
 }
 
+get_network_identifier() {
```
enxebre: Just minor thing: If we pass the output of $( terraform output vpc_cidr_block.ip  ) as an argument to get_network_identifier the we can test that function in test.sh

enxebre:
lgtm

tayzlor:
thanks!

PR #382, opened 2015-07-13
Refactoring Zookeeper peers looping so we DRY across roles

ravbaba:


roles/mesos/templates/zk.j2
```diff
@@ -1,14 +1,3 @@
...
-{%- for host in groups[mesos_zookeeper_group] -%}
-  {%- if not loop.last -%}
-    {%- set mesos_zk_url = mesos_zk_url + hostvars[host]['ansible_ssh_host'] + ':2181,' -%}
-  {%- endif -%}
-{%- if loop.last -%}
-  {%- set mesos_zk_url = mesos_zk_url + hostvars[host]['ansible_ssh_host'] + ':2181/mesos' -%}
-zk://{{ mesos_zk_url }}
-  {%- endif -%}
-{%- endfor -%}
+{% if zookeeper_peers_nodes is defined %}
```
enxebre: Is there anycase where zookeeper_peers_nodes is not defined? I think it should be always defined, otherwise this should raise an error.

ravbaba: You are right we don't need the condition

enxebre:
Looks good to me!

PR #391, opened 2015-07-15
Toggling of debug

ravbaba:


bootstrap/apollo-env.sh
```diff
@@ -14,3 +17,24 @@ if [[ -z "${color_start-}" ]]; then
...
+# if unset allow for existing selective logging of components
+case "${APOLLO_LOG}" in
+  "")
+    # Do nothing in this instance
+  ;;
+  0)
+    # Force minimal logging
+    echo "Forcing reduction of component logging"
+    unset TF_LOG
+    unset ANSIBLE_LOG
```
enxebre: This will break ansible_playbook_run -> ANSIBLE_LOG: unbound variable

bootstrap/apollo-env.sh
```diff
@@ -14,3 +17,23 @@ if [[ -z "${color_start-}" ]]; then
...
+# Change logging levels of called components at a global level
+# if unset allow for existing selective logging of components
+case "${APOLLO_LOG}" in
+  "")
+    # Do nothing in this instance
+  ;;
+  0)
+    # Force minimal logging
+    echo "Forcing reduction of component logging"
+    unset TF_LOG
```
enxebre: We still need to have a consitent behaviour here. So may be just set the variables to empty:
export ANSIBLE_LOG=
export TF_LOG=

Once we test it with this change lgtm.

PR #394, opened 2015-07-15
Installing jq on base image.

enxebre:


pmbauer:
rather than install ad-hoc client tooling on the servers, could we just put things like jq in a docker image that can be pulled on an as-needed basis?

tayzlor:
kinda like the docker idea keeps it clean, could use https://registry.hub.docker.com/u/realguess/jq/

enxebre:
Yup agree, the idea is to keep the base image as slim as posible moving components inside containers so same aproach can apply here. Thinking if it worth to move all similar tools in packer scripts into a custom tooling image instead of individual images as the one above.

tayzlor:
depends, i think a few of them need to be installed on the machine, but things like ruby / bundler etc could potentially be moved to some docker image to support the serverspecs. I'm not sure if we can move them all?

enxebre:
Closing as per https://github.com/Capgemini/Apollo/issues/414

PR #396, opened 2015-07-17
DCOS cassandra

enxebre:


roles/frameworks/templates/cassandra-config.j2
```diff
@@ -0,0 +1,7 @@
+{
```
tayzlor: should we expose the other config vars here or just leave it for now? https://github.com/mesosphere/universe/blob/version-1.x/repo/packages/C/cassandra/0/config.json

roles/frameworks/defaults/main.yml
```diff
@@ -0,0 +1,9 @@
+---
+# defaults file for frameworks
+frameworks_dcos_cli_image: capgemini/dcos-cli
+frameworks_zk_master_peers: "zk://{{ zookeeper_peers_nodes }}/mesos"
+frameworks_mesos_master_url: "http://{{ ansible_ssh_host }}:5050"
+frameworks_marathon_url: "http://{{ ansible_ssh_host }}:8080"
```
tayzlor: should this use ```{{ marathon_port }}```

roles/frameworks/defaults/main.yml
```diff
@@ -0,0 +1,9 @@
+---
+# defaults file for frameworks
+frameworks_dcos_cli_image: capgemini/dcos-cli
+frameworks_zk_master_peers: "zk://{{ zookeeper_peers_nodes }}/mesos"
+frameworks_mesos_master_url: "http://{{ ansible_ssh_host }}:5050"
```
tayzlor: Should use ```{{ mesos_master_port }}```

tayzlor:
:+1: really like this PR!

wallies:
very nice :+1:

roles/frameworks/defaults/main.yml
```diff
@@ -0,0 +1,11 @@
+---
+# defaults file for frameworks
+frameworks_dcos_cli_image: capgemini/dcos-cli
+frameworks_zk_master_peers: "zk://{{ zookeeper_peers_nodes }}/mesos"
+frameworks_mesos_master_url: "http://{{ ansible_ssh_host }}:{{ mesos_master_port }}"
+frameworks_marathon_url: "http://{{ ansible_ssh_host }}:{{ marathon_port }}"
+
+frameworks_cassandra_enabled: true
```
wallies: are we going to enable this by default. all frameworks except marathon should be disabled by default i think

enxebre: ups no, was testing

roles/frameworks/templates/cassandra-config.j2
wallies: even expose the minimum required configuration values would be nice with the default in the config.json e.g.
"framework","cluster-name","zk-timeout-ms","node-count","seed-count","health-check-interval-seconds","bootstrap-grace-time-seconds","data-directory","resources"

tayzlor: can be a follow-up issue

PR #407, opened 2015-07-21
Moving chronos into dcos-cli and update docs.

enxebre:


tayzlor:
nice! looks good to me

roles/frameworks/vars/cassandra.yml
```diff
@@ -0,0 +1,2 @@
+frameworks_cassandra_enabled: false
```
tayzlor: exceptionally minor nitpick, but the plural here doesn't make loads of sense (but realise you are namespacing it after the role name) it would read better / make more sense as ```framework_cassandra_enabled```

PR #408, opened 2015-07-21
Fixing kernel-upgrade.sh script

ravbaba:


packer/scripts/ubuntu/kernel-upgrade.sh
```diff
@@ -50,12 +50,12 @@ then
   sleep 2s
 
   echo "Installing the packages..."
-  dpkg -i "$DEB"
+  sudo dpkg -i $DEB
```
wallies: add this DEBIAN_FRONTEND=noninteractive, so it would be sudo DEBIAN_FRONTEND=noninteractive dpkg -i $DEB

packer/scripts/ubuntu/kernel-upgrade.sh
```diff
@@ -66,7 +66,7 @@ then
...
```
wallies: as above

enxebre:
Labeling this as needs work as there are some more issues that needs to be addressed around the kernel upgrade.

PR #409, opened 2015-07-21
Upgrade terraform.

enxebre:


tayzlor:
does this affect boostrap/common.sh - we check for terraform version 0.5.0 ?

tayzlor:
There's also ```local requirement_version_string=${1:-0.5.0}``` should we update?

enxebre:
That's just the default value for the function which does not apply as we are passing the the minimum version as a parameter. Can update it too though

PR #413, opened 2015-07-22
change scripts to reference docker-py 1.3.0

wallies:


packer/scripts/ubuntu/base.sh
```diff
@@ -12,5 +12,3 @@ sudo apt-get update -y
 
 sudo apt-get -y install git curl libcurl3 bridge-utils bundler default-jre-headless unzip wget python-setuptools python-protobuf cgroup-bin ruby1.9.1
 sudo easy_install pip
-
-sudo pip install docker-py==1.1.0
```
enxebre: should we just keep this one and remove all the other ones?

wallies: i moved it to install_docker.sh as that's where it belongs. removed from the roles and added spec tests for packer

enxebre: nice

enxebre: I just tested and it breaks.Needs more testing https://github.com/ansible/ansible-modules-core/issues/1227

enxebre: We need to test this now that https://github.com/Capgemini/Apollo/pull/423 got merged.

enxebre: Seems that's will be fix in next ansible version https://github.com/ansible/ansible-modules-core/issues/1707

wallies:
this now works and passes the dockerbench tests. not sure why it didnt pass apollo deploy but i think thats a seperate issue

enxebre:
As per the ticket in the comment above https://github.com/ansible/ansible-modules-core/issues/1707#issuecomment-122617536 dont think this will work until ansible 2.0 get released as it's not compatible with new versions docker remote api.
This will require rebuild the images.
For testing it without rebuild you can install the new version of docker-py from an ansible task so will override the docker-py version of the images.

wallies:
it was included in 1.9.3. I upgraded to 1.9.3 and upgraded to docker-py 1.3.1 via the ansible tasks, then logged it to a few of the machines and ran dockerbench.
[PASS] 5.11 - Limit memory usage for container
[PASS] 5.12 - Set container CPU priority appropriately
I would say we merge this and the other PR relating to setting docker memory and then rebuild the images

PR #424, opened 2015-07-23
Bump slave mem to 1532MB in vagrant.yml

zytek:
Currently Chronos fails to start under Marathon because when
slave goes up it has less then 512MB mem free

(at least on OSX Yosemite + VirtualBox 4.3.30 + vagrant 1.7.4)


enxebre:
true that, thanks!

Issue #416, opened 2015-07-22
Sort out solution for testing frameworks deployed by dcos-cli


enxebre:
When testing that services are running we run serverspecs from inside every host pointing to localhost. As frameworks deployed by dcos-cli are deployed arbitrary into any salve we need to put in place a more dynamic testing solution. e.g Point to elb from the serverspecs

PR #426, opened 2015-07-23
Deployment per PR

enxebre:
Increase dockerbench threshold as we need to wait for https://github.com/Capgemini/Apollo/pull/413 for fixing memory limit container warnings.

Remove chronos specs as we need to address a solution for testing frameworks now https://github.com/Capgemini/Apollo/issues/416

Provides overridable name for elb. This could just be removed in terraform 0.6.2 https://github.com/hashicorp/terraform/blob/master/CHANGELOG.md as name becomes optional

Provides ability to parallel deployments per PR so we can decide now if want to set this up.


wercker.yml
```diff
@@ -42,6 +44,17 @@ build:
         cwd: bootstrap/tests
         code: |
           /bin/bash test.sh
+    - capgemini/apollo-deploy:
+        cloud: aws-public
+        artifact_name: capgemini/apollo-ubuntu-14.04-amd64
+        artifact_version: 6
+        terraform_version: 0.5.3
```
tayzlor: probably should bump this inline with the code we merged the other day in common.sh

tayzlor:
lgtm

enxebre:
I'll keep this out of the wercker build step for now.

PR #427, opened 2015-07-27
Changed artifact version

ravbaba:
Changed to a new artifact version with hvm virtualization

### DIFF
--- a/terraform/aws-public/mesos-slaves.tf
+++ b/terraform/aws-public/mesos-slaves.tf
@@ -7,8 +7,16 @@ resource "atlas_artifact" "mesos-slave" {
 
 /* Mesos slave instances */
 resource "aws_instance" "mesos-slave" {
+  /*
+     We had to hardcode the amis list in variables.tf file creating amis map because terraform doesn't
+     support interpolation in the way which could allow us to replaced the region dinamically.
+     We need to remember to update the map every time when we build a new artifact on atlas.
+     Similar issue related to metada_full is mentioned here:
+     https://github.com/hashicorp/terraform/issues/732
+  */
+
   instance_type     = "${var.instance_type.slave}"
-  ami               = "${replace(atlas_artifact.mesos-master.id, concat(var.region, ":"), "")}"
+  ami               = "${lookup(var.amis, var.region)}"
   count             = "${var.slaves}"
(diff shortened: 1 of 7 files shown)


terraform/aws-public/mesos-slaves.tf
```diff
@@ -8,7 +8,7 @@ resource "atlas_artifact" "mesos-slave" {
 /* Mesos slave instances */
```
enxebre: Can we add a comment here exaplaining why is this being harcoded and linking to terraform issue?

enxebre:
This needs to wait for having vagrant image with same version of ubuntu and linux kernel

terraform/aws-public/mesos-slaves.tf
tayzlor: could we not do some kind of map here in the variables?

``
amis = {
 eu-west-1 => 'ami-123',
 us-east-1 => 'ami-456',
}
```
etc..

and then use that var to replace on the end..

```
 "${atlas_artifact.mesos-master.metadata_full.region-${var.amis.${region}}}"
```
or whatever is is

Comment (enxebre) [inline terraform/aws-public/mesos-slaves.tf]: Dont think that interpolation is supported by terraform. ->  "${atlas_artifact.mesos-master.metadata_full.region-${var.amis.${region}}}"

Think we can harcode it as ${atlas_artifact.mesos-master.metadata_full.region-eu-west-1}
or create a static map region -> ami-id
Then just ami = lookup(var.amis, var.region)

Static mapping would need to be updated every time we build images and the atlas_artifact resource shoudl be removed as it would be useless if we go for second option.

Comment (ravbaba) [inline terraform/aws-public/mesos-slaves.tf]: As enxebre mentioned the interpolation is not supported by terraform. I tested it with terraform version 0.6.1 (the newest one). So the code:
```
varname = "atlas_artifact.mesos-master.metadata_full.region-${var.region}”

ami = ${varname}
```
and my version mention in the [terraform ticket](https://github.com/hashicorp/terraform/issues/732): 

```
ami = "${lookup(atlas_artifact.mesos-master.metadata_full, concat("region-", var.region))}"
```
don't work.

enxebre:
This change would need to be applied to aws provider too.

wallies:
this lgtm

PR #430, opened 2015-07-28
Updated packer image for vagrant

ravbaba:
I updated only packer image for vagrant because rest of them didn't required any changes.


tayzlor:
thanks!

PR #435, opened 2015-07-29
Reverted bastion host to use stock Ubuntu AMI

ravbaba:


.gitignore
```diff
@@ -16,4 +16,4 @@ vendor
 inventory/terraform.py
 tests/spec/*/*runtime_spec.rb
 contrib-plugins/*
-
+terraform/aws/.terraform
```
enxebre: can we do terraform/*/.terraform ?

terraform/aws/bastion-server.tf
```diff
@@ -3,10 +3,21 @@ resource "aws_key_pair" "deployer" {
   public_key = "${file(var.key_file)}"
 }
 
+/*
+   Terraform module to get the current set of publicly available ubuntu AMIs.
+   https://github.com/terraform-community-modules/tf_aws_ubuntu_ami
+*/
```
enxebre: can we set this module name more specific like "ami_bastion"

Issue #392, opened 2015-07-15
Move zookeeper to a docker container


enxebre:
We should be able to move zookeeper out to a docker container so keeping the base image minimal
https://registry.hub.docker.com/search?q=zookeeper&searchfield=

enxebre:
https://github.com/Capgemini/Apollo/pull/441

Issue #129, opened 2015-04-30
  Move consul out to a docker container


wallies:
https://github.com/progrium/docker-consul

keithchambers:
FFIW we had no-so-great luck with Consul in Docker and moved it out.

tayzlor:
Thanks for the feedback Keith! Any particular issues you were facing with the Docker setup?

keithchambers:
ARP caching, restarting Docker takes it down, and just way too much code around making it all work. In the end it was far simpler and more reliable to run Consul outside a container.

wallies:
Keith our main motivation for moving this out to a container is like other components, so we have very minimal on our base image. Eventually we aim to have nothing on the base image, so we can use public shared images like on the nat server.

enxebre:
There are new images for consul agent and master provided by gliderlabs https://registry.hub.docker.com/repos/gliderlabs/

wallies:
this is now supported in coreos devel branch

PR #438, opened 2015-07-30
Dnsmasq running inside container.

enxebre:
This needs to wait for a packer build for being merged.
We might want wait for https://github.com/Capgemini/Apollo/issues/129 and https://github.com/Capgemini/Apollo/issues/392


roles/marathon/templates/marathon-wait-for-listen.sh.j2
```diff
@@ -1,10 +1,10 @@
 #!/bin/bash
 
-MAX_SECONDS=60
+MAX_SECONDS=120
```
enxebre: Had to increase this time out, needs more investigation on why this happened.

enxebre:
Just need to wait for https://github.com/Capgemini/Apollo/pull/441 to merged now as the ami ids here include changes in that PR

PR #444, opened 2015-07-31
Downgrade docker version from 1.7.1 to 1.7.0

ravbaba:
This is a temporary solving to fix the issue. It is gonna be fixed in the new docker version.


wallies:
Can we move to the new docker official repos that install docker-engine instead if lxc-docker
https://blog.docker.com/2015/07/new-apt-and-yum-repos/

ravbaba:
I think you are right but the point was to downgrade the version to 1.7.0 because it fixes the issue. It is only temporary solution until it is fixed in new docker version 1.8. The repos which install docker-engine gives you version 1.7.1. We should definitely do that when they fix it and release new version 1.8 what I think it's going to happen in next week.

enxebre:
yep, let's move to the new repos once the new version after 1.7.1 get released.
https://github.com/Capgemini/Apollo/issues/445

PR #447, opened 2015-08-03
Set -resync flag for registrator reregistering all services

ravbaba:
This allows registrator and the service registry to get back in sync if they fall out of sync


wallies:
this lgtm, is there anything blocking this?

PR #448, opened 2015-08-03
Consul on docker.

enxebre:


roles/consul/tasks/config.yml
```diff
@@ -0,0 +1,43 @@
+# tasks file for consul
+- name: remove consul override
```
tayzlor: is this still needed?

roles/consul/tasks/config.yml
```diff
@@ -0,0 +1,43 @@
...
+  file:
+    path: /etc/init/consul.override
+    state: absent
+
+- service: name=consul state=stopped enabled=no
```
tayzlor: needed?

wallies:
@enxebre this might be worth another go with the new version on consul

tayzlor:
Closing in favour of #585

PR #449, opened 2015-08-04
Moved to the new docker official repos that install docker-engine ins…

ravbaba:
We need to wait for the new docker release 1.8, after that and rebuilding aws and vagrant images and testing we can merge it.


packer/scripts/ubuntu/install_docker.sh
```diff
@@ -5,21 +5,14 @@ set -o pipefail
 sudo apt-get -y update
 sudo apt-get install -y linux-image-extra-$(uname -r)
```
enxebre: The point of doing this manually rather than just "curl -sSL https://get.docker.com/ | sh" is to have the ability to lock to a given version and we are removing that here.
Can not we do "sudo apt-get install -y mesos=${MESOS_VERSION}" the something like "it { should be_installed.by('apt').with_version()"

tayzlor:
Since we are doing a rebuild of the images here might be worth adding in https://github.com/Capgemini/Apollo/issues/457 to this PR

enxebre:
Images have been rebuilt already. Lets get this in for the sake of stability and create a follow up for testing and merge #457 in insolation

tayzlor:
this good to go then? what we waiting for?

tayzlor:
thanks!

PR #461, opened 2015-08-25
Add line to enable memory and swap limits

wallies:
By default, the debian kernels are not started with cgroup_enable=memory and awapaccount=1 parameters. As a result, cgroup does not correctly enforce memory and swap limits on a container. This PR enables this


tayzlor:
needs rebased

Issue #463, opened 2015-08-25
docker should have_volume tests failing.


enxebre:
Using docker 1.8.1 docker should have_volume tests are failing.

wallies:
working and moving forward will track here #583

PR #464, opened 2015-08-25
Upgrade serverspecs and temporally comment docker volume tests.

enxebre:
Upgrades serverspec, specinfra and temporally comment out should have_volume as it does not work with docker 1.8.1
Probably because changes introduces due to volume drivers from experimental to master
https://github.com/Capgemini/Apollo/issues/463


tayzlor:
lgtm

PR #466, opened 2015-08-31
Add changes for docker partition on aws

wallies:


bootstrap/aws-public/util.sh
```diff
@@ -18,6 +18,27 @@ EOF
   popd
 }
 
+mount_docker_volume_script() {
```
tayzlor: any reason to only do this on AWS? 

We could generalise this function somewhere and call it from each rather than copy/pasting it across 2 files

Can this not be done from ansible? http://docs.ansible.com/ansible/mount_module.html

PR #468, opened 2015-09-01
Udacity secure elb

pmbauer:
This secures the aws load balancer, only allowing TLS traffic, terminating at the load balancers.
Only 'public' -tagged services in consul will be exported; that means marathon, mesos etc aren't exposed to the world.
We are using udacity's version of haproxy-consul because we need to use v0.10.0 of consul-template
https://github.com/udacity/haproxy-consul


enxebre:
Looks good to me.
Thanks!

wallies:
closing as we no longer use haproxy. will revisit this when traefik letsencrypt support lands

PR #471, opened 2015-09-01
Udacity add weave scope disable flag

pmbauer:
weave scope is a massive resource-hog, often taking up constant 20%+ cpu on slaves and masters.
As more containers are launched on the cluster, the scope UI doesn't scale well and the utility of weave scope doesn't justify its cost.  We needed a flag to turn it off.


tayzlor:
thanks!

PR #472, opened 2015-09-01
Udacity rebuild container flags

pmbauer:
allows the operator to selectively taint running containers and just run the idempotent bootstrap/apollo-launch.sh, (e.g. changing config files for haproxy, etc)


enxebre:
Nice! I'll create ticket as a follow up for applying the same approach to the other containers i.e mesos master, mesos slave, dnsmasq, zookeeper.
We could also add a global variable rebuild_all_containers so then per container:
when: marathon_rebuild_container OR rebuild_all_containers
Thanks!!
LGTM.

PR #475, opened 2015-09-02
Update vpn.md

boostrack:


tayzlor:
Thanks for the PR. The doc you add is actually supposed to be getting handled here :

https://github.com/Capgemini/Apollo/blob/master/bootstrap/aws/util.sh#L72-L75

it could be that's not quite working though...

boostrack:
Yes thanks. I also figured that as you pointed but if from a ./apollo-launch on first run you don't configure ovpn and following the steps from vpn guide one faces this behavior

tayzlor:
ah I see what you mean - good point, thanks for the doc update

PR #320, opened 2015-06-19
Azure provider

tayzlor:
Opening this up but it's still a WIP.

On the TODO - 

- [ ] Build the Azure image from Packer using https://github.com/msopentech/packer-azure and test it
- [ ] Complete the terraform scripts to reference built image and test fully the provisioning of instances in azure
- [ ] Update the terraform dynamic inventory to handle Azure
- [ ] Add docs around getting started on Azure

PR #493, opened 2015-10-19
Azure provider

ravbaba:
Still need more work

### DIFF
--- a/bootstrap/azure/config-default.sh
+++ b/bootstrap/azure/config-default.sh
@@ -6,8 +6,7 @@ export ATLAS_INFRASTRUCTURE=${ATLAS_INFRASTRUCTURE:-capgemini/apollo}
 
 export TF_VAR_azure_settings_file=${TF_VAR_azure_settings_file:?"Need to set TF_VAR_azure_settings_file non-empty"}
 export TF_VAR_username=${TF_VAR_username:?"Need to set TF_VAR_username non-empty"}
-export TF_VAR_password=${TF_VAR_password:?"Need to set TF_VAR_password non-empty"}
-#export TF_VAR_ssh_key_thumbprint=${TF_VAR_ssh_key_thumbprint:?"Need to set TF_VAR_ssh_key_thumbprint non-empty"}
+export TF_VAR_ssh_key_thumbprint=${TF_VAR_ssh_key_thumbprint:?"Need to set TF_VAR_ssh_key_thumbprint non-empty"}
 
 # Overrides default folder in Terraform.py inventory.
 export TF_VAR_STATE_ROOT="${APOLLO_ROOT}/terraform/${APOLLO_PROVIDER}"
--- a/bootstrap/azure/config-default.sh
+++ b/bootstrap/azure/config-default.sh
@@ -5,8 +5,11 @@ export ATLAS_TOKEN=${ATLAS_TOKEN:?"Need to set ATLAS_TOKEN non-empty"}
 export ATLAS_INFRASTRUCTURE=${ATLAS_INFRASTRUCTURE:-capgemini/apollo}
 
 export TF_VAR_azure_settings_file=${TF_VAR_azure_settings_file:?"Need to set TF_VAR_azure_settings_file non-empty"}
+export TF_VAR_user=${TF_VAR_user:?"Need to set User non-empty"}
 export TF_VAR_username=${TF_VAR_username:?"Need to set TF_VAR_username non-empty"}
-export TF_VAR_ssh_key_thumbprint=${TF_VAR_ssh_key_thumbprint:?"Need to set TF_VAR_ssh_key_thumbprint non-empty"}
+
+export TF_VAR_key_file=${TF_VAR_key_file:-$HOME/.ssh/apollo_azure_rsa.pfx}
+export TF_VAR_key_name=${TF_VAR_key_name:-apollo}
 
 # Overrides default folder in Terraform.py inventory.
 export TF_VAR_STATE_ROOT="${APOLLO_ROOT}/terraform/${APOLLO_PROVIDER}"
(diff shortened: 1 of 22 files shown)


bootstrap/common.sh
```diff
@@ -3,6 +3,9 @@
 # Util functions cloud reusable.
 APOLLO_ROOT=$(dirname "${BASH_SOURCE}")/..
 DEFAULT_CONFIG="${APOLLO_ROOT}/bootstrap/${APOLLO_PROVIDER}/${APOLLO_CONFIG_FILE-"config-default.sh"}"
+# It must be change before mergeing with capgemini master branch
+DYNAMIC_INVENTORY="https://raw.githubusercontent.com/ravbaba/terraform.py/azure-provider/terraform.py"
```
ravbaba: The url should be change back to https://raw.githubusercontent.com/Capgemini/terraform.py/master/terraform.py after merging my PR for terraform.py repo

enxebre:
Is this overlapping with https://github.com/Capgemini/Apollo/pull/320 ?

terraform/azure/mesos-slaves.tf
```diff
@@ -0,0 +1,92 @@
...
+    private_port = 8080
+  }
+
+  endpoint {
+    name         = "consul"
+    protocol     = "tcp"
+    public_port  = 8500
+    private_port = 8500
+  }
+
```
wallies: might need some more endpoints for consul here as these are the ports that consul uses
Client Addr: 127.0.0.1 (HTTP: 8500, DNS: 8600, RPC: 8400)
Cluster Addr: 192.168.1.43 (LAN: 8301, WAN: 8302)

enxebre:
What's the current state of this? still WIP? I haven't made an Azure deployment but code looks pretty good to me.
This needs Rebase on master though.

ravbaba:
I am gonna create a new PR with a different approach. Because I encountered some difficulties using terraform with azure so I decided to do it with Azure Resource Manager template using azure cli.

https://azure.microsoft.com/en-gb/documentation/articles/resource-group-template-deploy/#deploy-with-azure-cli-for-mac-linux-and-windows

I am going to build an inventory file base on the example.

https://github.com/ansible/ansible/blob/devel/contrib/inventory/windows_azure.py

Anyway we should keep the PR just in case if terraform will have better support for Azure. At the moment there are not support for ARM.

enxebre:
Setting this PR to blocked for now then.

terraform/azure/security-groups.tf
```diff
@@ -0,0 +1,32 @@
+/* Default security group */
+resource "azure_security_group" "default" {
+  name     = "default-apollo-mesos"
+  location = "${var.region}"
+}
+
+resource "azure_security_group_rule" "all-inbound" {
```
tayzlor: There seems to be some special constants described here - https://azure.microsoft.com/en-us/documentation/articles/virtual-networks-nsg/ such as INTERNET / VIRTUAL_NETWORK - might be worth double checking the docs

terraform/azure/security-groups.tf
```diff
@@ -0,0 +1,32 @@
...
+  name                       = "all-inbound-access-rule"
+  security_group_names       = ["${azure_security_group.default.name}"]
+  type                       = "Inbound"
+  action                     = "Allow"
+  priority                   = 100
+  source_address_prefix      = "*"
+  source_port_range          = "*"
+  destination_address_prefix = "*"
+  destination_port_range     = "*"
+  protocol                   = "TCP"
```
tayzlor: should be * to allow consul over UDP

tayzlor:
Removing blocked label - we have a way forward with https://github.com/hashicorp/terraform/pull/3695

terraform/azure/bin/ovpn-client-config
```diff
@@ -0,0 +1 @@
+ssh -t "ubuntu@$(terraform output bastion.ip)" sudo docker run --volumes-from ovpn-data --rm gosuri/openvpn ovpn_getclient "${1}" > "${1}-apollo.ovpn"
```
ravbaba: I think we can move the code to a common folder because it exact copy of that which we use for aws

bootstrap/azure/util.sh
```diff
@@ -0,0 +1,96 @@
...
+  read -p "Do you want to start the VPN and setup a connection now (y/n)?" yn
+    case $yn in
+        [Yy]* ) ovpn_start;ovpn_client_config; break;;
+        [Nn]* ) exit;;
+        * ) echo "Please answer y or n.";;
+    esac
+  done
+}
+
+ansible_ssh_config() {
```
ravbaba: The code should be move to a common library because it is a copy of existing one which we use for aws

bootstrap/azure/util.sh
```diff
@@ -0,0 +1,96 @@
...
+}
+
+apollo_down() {
+  pushd "${APOLLO_ROOT}/terraform/${APOLLO_PROVIDER}"
+    terraform destroy -var "azure_settings_file=${TF_VAR_azure_settings_file}" \
+      -var "region=${TF_VAR_region}"
+  popd
+}
+
+ovpn_start() {
```
ravbaba: The code should be move to a common library because it is a copy of existing one which we use for aws

bootstrap/azure/util.sh
```diff
@@ -0,0 +1,96 @@
...
+  pushd "${APOLLO_ROOT}/terraform/${APOLLO_PROVIDER}"
+    echo "... initialising VPN setup" >&2
+    /bin/sh -x bin/ovpn-init
+    /bin/sh -x bin/ovpn-start
+  popd
+}
+
+ovpn_client_config() {
```
ravbaba: The code should be move to a common library because it is a copy of existing one which we use for aws

bootstrap/azure/config-default.sh
```diff
@@ -0,0 +1,32 @@
...
+export TF_VAR_key_name=${TF_VAR_key_name:-apollo}
+
+# Overrides default folder in Terraform.py inventory.
+export TF_VAR_STATE_ROOT="${APOLLO_ROOT}/terraform/${APOLLO_PROVIDER}"
+
+export ANSIBLE_SSH_ARGS="-F ${APOLLO_ROOT}/terraform/${APOLLO_PROVIDER}/ssh.config -q"
+
+# Terraform mappings needs to be statically passed as -var parameters
+# so no really needed to export them. Exporting for consitency.
+export TF_VAR_atlas_artifact_master=${TF_VAR_atlas_artifact_master:-apollo-ubuntu-14.04-amd64-1446553949}
```
tayzlor: are we using atlas artifacts here?

ravbaba: we use azure images which are kept on azure storage here but I use the same variables for that because they are hardcoded in common.sh file in terraform_apply function so if I remove them I will get an error. Probably we could use here a default option to avoid the problem.

bootstrap/azure/util.sh
wallies: @ravbaba  can you raise a issue to do this. we can then look to merge this and iterate for improvements

PR #503, opened 2015-10-29
Add prometheus monitoring for nodes / containers

tayzlor:
Opening this up for early sight / feedback. Still on the todo list - 

- [x] Add consul health check for node-exporter (Added in https://github.com/Capgemini/Apollo/commit/9ab5cd094df75a786601f317d490506660aea6dc) 
- [x] Add consul health check for mesos-exporter (Added in https://github.com/Capgemini/Apollo/commit/ba1ad7b342d68df9972191a8ed2a633a3f9004fc)
- [x] Add marathon/consul health check for prometheus Docker container (Added in https://github.com/Capgemini/Apollo/commit/741b0b73d0f02df8f859fc452ddc83d93e7733ff)
- [ ] Install and configure https://hub.docker.com/r/prom/promdash/ Docker container (#506)
- [ ] Install and configure https://hub.docker.com/r/prom/alertmanager/ Docker container (#507)
- [ ] Add a shared data volume across the slave instances so that we can place the storage for prometheus on these volumes so if the container dies and gets restarted elsewhere in the cluster we don't lose the data. How do you setup and attach a volume across multiple instances? https://www.terraform.io/docs/providers/aws/r/volume_attachment.html looks like a 1-to-1 (#508)
- [x] Add a config switch to enable / disable prometheus monitoring (Added in https://github.com/Capgemini/Apollo/commit/54d3bf95daf31bd87df1940bb3d00e25e16df82c)

As an aside we should improve the HAproxy configuration for the backend services so we can access prometheus / promdash correct through the load balancer. We might also want to think about fronting it with some Auth - I guess these could both come in follow-ups / separate tickets. There is already some improved haproxy config here (that only surfaces 'public' services) https://github.com/Capgemini/Apollo/pull/468/files 

At the moment i've only tested this with aws-public - needs to be further tested on VPC as well.


roles/weave/defaults/main.yml
```diff
@@ -27,4 +27,4 @@ weave_launch_peers: "
 weave_docker_opts: "--bridge=weave --fixed-cidr={{ weave_docker_subnet }} --dns 172.17.42.1 --dns 8.8.8.8 --dns-search service.{{ consul_domain }}"
 weave_scope_url: https://github.com/weaveworks/scope/releases/download/latest_release/scope
 weave_scope_dest: /usr/local/bin/scope
-weave_scope_enabled: True
\ No newline at end of file
+weave_scope_enabled: true
```
tayzlor: this should be false - i'll correct it.

roles/dcos_cli/templates/prom-node-exporter.json.j2
```diff
@@ -0,0 +1,15 @@
+{
```
wallies: How does this get run by dcos-cli?

tayzlor: Have a look @ ```roles/dcos_cli/tasks/apps.yml```

tayzlor: my bad - that file needs removed

roles/dcos_cli/tasks/apps.yml
```diff
@@ -0,0 +1,36 @@
...
+    dest: "/etc/marathon/{{ item }}.json"
+    mode: 0755
+  sudo: yes
+  tags:
+    - "{{ item }}"
+    - dcos_cli
+  with_items:
+    - "{{ dcos_cli_apps_list }}"
+
+- name: "add marathon app"
```
enxebre: could add "delete marathon app" task here

pmbauer:
This is really neat! We'll revisit our monitoring decisions at Udacity, but in the mean time, we added datadog's agent across our cluster (native mesos, marathon, haproxy, docker, and consul support).  We'd send a pull request upstream for a global flag to opt in/out of Prometheus for the near term.

tayzlor:
Cool - we could probably build that flag into this PR - i'll add a todo

tayzlor:
wonder if we should split out 4 & 5 on the todo for later or do them as part of this? thoughts?

wallies:
yeh 4,5,6,7 feel like they should be split into seperate tasks. Maybe include 7 as part of this PR.

tayzlor:
I've added a config switch around the main prometheus container in https://github.com/Capgemini/Apollo/commit/741b0b73d0f02df8f859fc452ddc83d93e7733ff . What that didn't do is add switches around the node-exporter and the mesos-exporter. Do we think that's necessary?

pmbauer:
@tayzlor wonderful thanks; yes we'd like to opt-in to the [node|mesos]-exporter since we already have an equivalent daemons scraping metrics for our monitoring provider

tayzlor:
@pmbauer - OK, thanks for the feedback, i'll rejig and add a global switch for all prometheus related items.

tayzlor:
With my latest commit - I think we could mark this particular PR as feature complete (pending code review) and roll the other items into separate individual issues. 

I just need to do some further testing using the AWS VPC setup.

tayzlor:
I've road tested this on the AWS VPC setup and it appears to be working well...

wallies:
@tayzlor can you make sure you raise issues for the other three issues and make a label for the first two for prometheus.

tayzlor:
@wallies new issues created for the sub tasks

enxebre:
We need to sort this out https://github.com/Capgemini/Apollo/issues/513
and run cadvisor through the weave proxy as we are relying on the container ip being reachable from anywhere for the consul check here.

wallies:
@enxebre i think this should be a seperate task

wallies:
this lgtm and tested on aws

wallies:
closing this in favour of #526

PR #504, opened 2015-10-30
Weave upgrade 1.2.0, fastadapath and respecting docker0 as default bridge

enxebre:
This provides a pretty basic weave 1.2.0 setup so using fastdatapath and respecting the docker0 bridge as the default docker bridge.

**Known issues:**
When running a new service on the cluster Registrator is storing in the first place the docker ip instead of the weave one. If Registrator get restarted it stores the weave ip.
This hopefully will be fixed once this gets it https://github.com/weaveworks/weave/issues/1549
So we can rebuild the images then.

Once docker 1.9 gets released weave/registrator/consul ecosystem will be broken unless this gets support https://github.com/weaveworks/weave/issues/1601

**TODOS:**
To provide Application/environment isolation via weave subnets.

**Future roadmap:**
To investigate and implement a more flexible/pluggable networking system for Apollo once Docker 1.9 with libnetwork support get released https://github.com/docker/libnetwork and mesos net modules becomes stable.
See https://github.com/docker/libnetwork/blob/master/docs/overlay.md
See https://github.com/projectcalico/calico-docker/blob/master/docs/getting-started/libnetwork/Demonstration.md
Once we move to mesos 0.25.0:
See http://mesos.apache.org/documentation/latest/networking-for-mesos-managed-containers/
See https://github.com/mesosphere/net-modules

### DIFF
--- a/site.yml
+++ b/site.yml
@@ -28,8 +28,8 @@
     - docker
     - weave
     - dnsmasq
-    - registrator
     - cadvisor
+    - registrator
 
 - hosts: mesos_masters
   roles:
--- a/packer/scripts/common/install_weave.sh
+++ b/packer/scripts/common/install_weave.sh
@@ -3,6 +3,6 @@ set -eux
 set -o pipefail
 
 sudo wget -O /usr/local/bin/weave \
-    https://github.com/weaveworks/weave/releases/download/${WEAVE_VERSION}/weave
+    https://github.com/weaveworks/weave/releases/download/v${WEAVE_VERSION}/weave
 sudo chmod a+x /usr/local/bin/weave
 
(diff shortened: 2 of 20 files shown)


site.yml
```diff
@@ -28,8 +28,8 @@
     - docker
     - weave
     - dnsmasq
-    - registrator
     - cadvisor
+    - registrator
```
tayzlor: just wondering the significance of the movement here? Wondering if that will affect what i've done over here https://github.com/Capgemini/Apollo/pull/504/files (where we get the cadvisor IP) - i'm guessing we might need to refactor that PR anyway based on this PR (or vice versa)

enxebre: Should be irrelevant as we are getting the ip in your PR via docker inspect, so all we need to do when merging both is making sure your "docker inspect"  goes through the weave proxy.

enxebre:
Seems like this would fix both know issues in one go https://github.com/weaveworks/weave/pull/1616/files

tayzlor:
the actual code changes in this PR look OK to me, so long as the functionality remains intact as we change approach.

PR #517, opened 2015-11-04
DOCKER_HOST to weave socket when running ansible.

enxebre:
This set the DOCKER_HOST env variable to be the weave socket at ansible runtime level so containers run by ansible docker module go through weave proxy.
The other possibility is to handle it at the task level an just set the socket as another variable same we are doing for the net, image, state or volumes. Personally I'd be in favour of the second approach so opening this for discussion.

### DIFF
--- a/group_vars/all
+++ b/group_vars/all
@@ -22,3 +22,4 @@ proxy_env:
   HTTP_PROXY: ''
   HTTPS_PROXY: ''
 http_proxy: ''
+docker_host: 'unix:///var/run/weave/weave.sock'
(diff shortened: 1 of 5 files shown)


group_vars/all
```diff
@@ -22,3 +22,5 @@ proxy_env:
   HTTP_PROXY: ''
   HTTPS_PROXY: ''
 http_proxy: ''
+
+weave_socket: 'unix:///var/run/weave/weave.sock'
```
tayzlor: i'd rename this either ```docker_socket``` or ```docker_host```

Issue #505, opened 2015-10-30
Upgrade docker version (currently 1.7.1) on bastion host


tayzlor:
The bastion in AWS is using an outdated method to install docker - 

```
"curl -sSL https://get.docker.com/ubuntu/ | sudo sh",
```

This wont retrieve anything newer than docker 1.7.1. There are different instructions for installing Docker (later versions) from the new repos - https://docs.docker.com/installation/ubuntulinux/

PR #518, opened 2015-11-04
Upgrading docker version on bastion host - issue #505

ravbaba:
fix for issue #505

### DIFF
--- a/terraform/aws/variables.tf
+++ b/terraform/aws/variables.tf
@@ -109,3 +109,8 @@ variable "amis" {
     us-west-2 = "ami-00657261"
   }
 }
+
+variable "docker_version" {
+  description = "Docker version"
+  default = "1.8.2-0~trusty"
+}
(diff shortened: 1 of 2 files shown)


terraform/aws/variables.tf
```diff
@@ -109,3 +109,8 @@ variable "amis" {
     us-west-2 = "ami-00657261"
   }
 }
+
+variable "docker_version" {
+  description = "Docker version"
+  default = "1.8.2-0~trusty"
```
wallies: @ravbaba  can we change this to 1.9

ravbaba: sure, let me test it first

enxebre: ideally should merge this one for the sake of consistency then tag so we got an snapshot with new weave approach plus the same version of docker across the cluster, then create a new pr for new version

tayzlor: Agreed - we could then build the 1.9 into https://github.com/Capgemini/Apollo/pull/520 on top of this change

PR #532, opened 2015-11-11
Fix #473 - Add rebuild container to mesos/zookeeper/cadvisor/dnsmasq

### LINKED ISSUE #473: Add rebuild_container to every container
Follow same approach than https://github.com/Capgemini/Apollo/pull/472 for the other containers i.e mesos master, mesos slave, dnsmasq, zookeeper.

We could also add a global variable "rebuild_all_containers" so then per container we could:
when: marathon_rebuild_container OR rebuild_all_containers

tayzlor:


roles/cadvisor/tasks/main.yml
```diff
@@ -1,4 +1,13 @@
 # tasks for running cadvisor
+- name: destroy old cadvisor container
+  when: cadvisor_rebuild_container
```
enxebre: Not only for this PR we should set a convention about using "| bool" filter or not with booleans as we have both cases now. The filter is safer in terms of a value passing though via automation e.g export variable used by bash script used by ansible

tayzlor: so we're saying we prefer this in all instances - 

```
when: cadvisor_rebuild_container|bool
```

roles/marathon/tasks/main.yml
```diff
@@ -9,7 +8,7 @@
     - marathon
 
 - name: create marathon artifact store directory
-  when: marathon_artifact_store_dir is defined and marathon_enabled | bool
+  when: marathon_artifact_store_dir is defined
```
wallies: do this marathon_artifact_store_dir is defined and marathon_artifact_store_dir != ''

tayzlor: Yeah there's a number of other places we have this though - do we want to create a separate issue for that? e.g. - 

```

/workspace/apollo/roles/consul/tasks/main.yml:
   57:   when: consul_join is defined
   71:   when: consul_join is defined

/workspace/apollo/roles/docker/tasks/main.yml:
   37:   when: http_proxy is defined and http_proxy != ''

/workspace/apollo/roles/marathon/tasks/main.yml:
   11:   when: marathon_artifact_store_dir is defined

/workspace/apollo/roles/mesos/defaults/main.yml:
   35: {% if mesos_master_quorum is defined %}

/workspace/apollo/roles/serverspec/tasks/main.yml:
   48:   when: test_role is defined and serverspec_run_tests

/workspace/apollo/roles/zookeeper/defaults/main.yml:
   10: 	{%- if zookeeper_host_list is defined -%}

/workspace/apollo/roles/zookeeper/templates/zoo.cfg.j2:
    6: {% if zookeeper_conf is defined %}
```

enxebre:
lgtm, just for the record: this removes ability to stop a "service" e.g marathon or cadvisor on second deploy.

roles/marathon/tasks/main.yml
wallies: yep raised #536

enxebre: then should just be 'marathon_artifact_store_dir != ''"" is defined is probably not need as they will be always defined in default.yml ?

tayzlor: maybe, lets discuss in #536

PR #534, opened 2015-11-11
kernel 3.19 images and overlay fs.

enxebre:
Setting the base images with kernel 3.19
Setting overlay as fs.
Paving the way for overlay network.

### DIFF
--- a/roles/docker/tasks/main.yml
+++ b/roles/docker/tasks/main.yml
@@ -15,7 +15,7 @@
     dest: /etc/default/docker
     state: present
     regexp: ^DOCKER_OPTS=.*--graph.*
-    line: 'DOCKER_OPTS=\"$DOCKER_OPTS --graph={{ docker_graph_dir }} --dns 172.17.0.1 --dns 8.8.8.8 --dns-search service.{{ consul_domain }} \"'
+    line: 'DOCKER_OPTS=\"$DOCKER_OPTS --storage-driver=overlay --graph={{ docker_graph_dir }} --dns 172.17.0.1 --dns 8.8.8.8 --dns-search service.{{ consul_domain }} \"'
   notify:
     - restart docker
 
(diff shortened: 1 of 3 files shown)


roles/docker/tasks/main.yml
```diff
@@ -15,7 +15,7 @@
     dest: /etc/default/docker
     state: present
     regexp: ^DOCKER_OPTS=.*--graph.*
-    line: 'DOCKER_OPTS=\"$DOCKER_OPTS --graph={{ docker_graph_dir }} --dns 172.17.0.1 --dns 8.8.8.8 --dns-search service.{{ consul_domain }} \"'
+    line: 'DOCKER_OPTS=\"$DOCKER_OPTS --storage-driver=overlay --graph={{ docker_graph_dir }} --dns 172.17.0.1 --dns 8.8.8.8 --dns-search service.{{ consul_domain }} \"'
```
wallies: do you want to do --storage-driver={{ docker_storage_driver }} instead and then in defaults put overlay

pmbauer:
I may have missed it, but is there a corresponding set of packer commits to upgrade the base images to 3.19?

roles/docker/tasks/main.yml
enxebre: Make sense I'll include that as part of a different PR

enxebre:
@pmbauer we are using now http://cloud-images.ubuntu.com/trusty/current/hwe-v/

pmbauer:
@enxebre what I mean is, I don't see any changes here https://github.com/Capgemini/Apollo/commits/master/packer

enxebre:
yeah there is no change in packer because the new base amis (upated in variables.tf) include the kernel 3.19 already :) http://cloud-images.ubuntu.com/trusty/current/hwe-v/trusty-server-hwe-v-cloudimg-amd64-kernel-info.txt

pmbauer:
Ah; sorry; I didn't realize 3.19 was coming in 14.04 LTE. Thanks.

PR #541, opened 2015-11-13
attach ebs to root partition.

enxebre:
At the moment we are attaching the ebs but then is not being mounted so it's not used.
This merge the ebs with the default ami mount root partition where we need more space as only the docker images makes it run out of space.


tayzlor:
image

tayzlor:
image

PR #542, opened 2015-11-13
add host entries for every machine

enxebre:


wallies:
what is this for?

enxebre:
so you can automate stuff bind to a given server ahead of time. e.g lock a marathon deploy to a specific server

wallies:
we have tags in terraform which give names to the servers
apollo-mesos-slave-${count.index}
so should be reachable with apollo-mesos-slave-0.consul.service

enxebre:
Those are aws tags, not marathon/consul tags, consul dns is just going to round roubin mesos-slave.consul.service among the running containers which does not necessary have to match a one to one relation with the host but it is not aware of mesos-slave.0.consul.service, we could tag the containers with a host index some how but I'd still like to have this at the machine level. We could improve with some discovery magic and dns though. This is just s simple solution for quickly satisfying a necessity

pmbauer:
I assume this is so you can use hostname constraints in a marathon app spec?
Are there scenarios where it's useful to hardcode a hostname constraint where, say, an attribute constraint wouldn't be more robust?

enxebre:
Basically I got an scenario where I want a on click bootstrap for jenkins running with the mesos plugin properly configured.
For achieving that I am running two containers one with jenkins as a service and another one with only data containing the jenkins configuration. For getting that working I need to bind the deployment of both containers to the same host, and I need something like zk://master1:2181,zk://master3:2181,zk://master3:2181/mesos to be resolvable as this is config necessary for the jenkins-mesos plugin. This does not need to be part of Apollo core at the moment so feel free to close this PR.

PR #543, opened 2015-11-15
Adding note about python

broomyocymru:
If python 3 used for pip ansible throws an error due to a print statement.


tayzlor:
thanks!

PR #546, opened 2015-11-26
traefik integration

tayzlor:
Opening this up, it's not quite complete, but maybe someone can take it over the line. Still on the todo -

- [ ] Ensure we have consul health checks for services (e.g. dnsmasq does not appear in consul now we have removed registrator
- [ ] Some further testing of the traefik integration


enxebre:
I really like this, I'll try to test it and get in ASAP as it's quite relevant for other PRs
If we got this it would take care of the discovery for the entry-point apps then we could rely on mesos dns or just docker overlay and docker dns for container to container conversation.

Thoughs?

tayzlor:
Superseeded by https://github.com/Capgemini/Apollo/pull/590

PR #547, opened 2015-11-27
Added promdash app

ravbaba:


roles/prometheus/tasks/main.yml
```diff
@@ -39,3 +39,8 @@
     - restart consul
   tags:
     - prometheus
```
enxebre: Does this mean we are running this db container on every host on the cluster? is that necessary?

group_vars/all
```diff
@@ -23,6 +23,7 @@ proxy_env:
   HTTPS_PROXY: ''
 http_proxy: ''
 prometheus_enabled: true
+promdash_enabled: true
```
enxebre: The flags on global works at the playbook level whereas yours is at the role level, we should probably try to be consistent

roles/dcos_cli/templates/promdash.json.j2
```diff
@@ -0,0 +1,39 @@
+{
+  "container": {
+    "type": "DOCKER",
+    "docker": {
+      "network": "BRIDGE",
+      "image": "{{ promdash_image }}",
+      "portMappings": [{"containerPort": 3000, "hostPort": 0, "protocol": "tcp"}]
```
wallies: im pretty sure we also need serviceport here, as we are running it on a bridge, also makes sure that the service all get the same port, when we scale.

enxebre: "Marathon does not make use of these assigned service ports itself. It is the responsibility of a separate component to make use of this information and set up load balancing accordingly. A sample script that does this is provided with Marathon"

Set servicePort is not going to make any different for us as we are not considering it in our discovery solution at the moment as we are not discriminating by port at the load balancer level and we are relying on container-ip (weave)/container-port for internal dns communication.

roles/prometheus/defaults/main.yml
```diff
@@ -7,3 +7,7 @@ prometheus_node_exporter_port: 9100
 prometheus_node_exporter_hostname: "{{ ansible_ssh_host }}"
 prometheus_node_exporter_consul_service_id: "{{ ansible_hostname }}:node-exporter:9100"
 prometheus_consul_dir: /etc/consul.d
+# promdash settings
+promdash_enabled: true
+promdash_database_url: "mysql2://promdash:promdash@promdash.mysql.service.consul/promdash"
+promdash_image: "ravbaba/promdash"
```
ravbaba: The image need to be uploaded to Capgemini docker hub account

roles/dcos_cli/vars/promdash.yml
```diff
@@ -0,0 +1,4 @@
+dcos_cli_app_promdash_enabled: "{{ promdash_enabled }}"
```
wallies: can this not be prometheus_enabled

roles/prometheus/defaults/main.yml
```diff
@@ -7,3 +7,7 @@ prometheus_node_exporter_port: 9100
 prometheus_node_exporter_hostname: "{{ ansible_ssh_host }}"
 prometheus_node_exporter_consul_service_id: "{{ ansible_hostname }}:node-exporter:9100"
 prometheus_consul_dir: /etc/consul.d
+# promdash settings
+promdash_enabled: true
```
wallies: why are we setting this to true here

roles/dcos_cli/vars/promdash.yml
ravbaba: you are right, we can use prometheus_enabled to promdash, I am going to change it

roles/prometheus/defaults/main.yml
ravbaba: @wallies - any preferences where you want me put it?

roles/dcos_cli/vars/promdash.yml
wallies: yeh i meant this bit "{{ promdash_enabled }}"

roles/prometheus/defaults/main.yml
wallies: Do we need it at all. It is already set in sites.yml

ravbaba:
I didn't manage to set up pre-configured dashboards. I found out how to export dashboards configuration and import it to empty dashboard later on but it requires to do some manual steps in promdash configuration, like adding prometheus server, a new folder and a new empty dashboard. How to export and import dashboards you can read here http://prometheus.io/docs/visualization/promdash/#pushing-dashboard-representations-as-json-via-http

roles/prometheus/defaults/main.yml
ravbaba: Sorry, I am confused, I can't find the setting in that file

wallies:
lgtm

roles/dcos_cli/templates/promdash.json.j2
```diff
@@ -0,0 +1,71 @@
...
+        }
+      },
+      "env": {
+        "SERVICE_NAME": "promdash",
+        "DATABASE_URL": "{{ promdash_database_url }}",
+        "RAILS_ENV": "production"
+      },
+      "cpus": {{ dcos_cli_app_promdash_cpus }},
+      "mem": {{ dcos_cli_app_promdash_mem }},
+      "ports": [3000],
```
enxebre: we don need to statically bind a port on the host

roles/dcos_cli/defaults/main.yml
```diff
@@ -8,8 +8,11 @@ dcos_cli_sources: '["https://github.com/Capgemini/universe/archive/version-1.x.z
 dcos_cli_frameworks_list:
   - cassandra
   - chronos
+# Command flag allows you to set up a command which you want to use to run an app.
+# To run one app use "command: app" and to run group of apps use "command: group".
 dcos_cli_apps_list:
-  - prometheus
+  - { name: prometheus, command: app }
+  - { name: promdash, command: group }
```
enxebre: Can we name this like "type"?

ravbaba: sure

PR #549, opened 2015-11-30
Elk docker

wallies:
added salesforce compability
- [x] Need to raise tickets for persistence of elasticsearch, better solution for logstash config and plugins


roles/dcos_cli/templates/kibana.json.j2
```diff
@@ -0,0 +1,39 @@
...
+      "portIndex": 0,
+      "initialDelaySeconds": 600,
+      "gracePeriodSeconds": 10,
+      "intervalSeconds": 30,
+      "timeoutSeconds": 120,
+      "maxConsecutiveFailures": 10
+    }
+  ],
+  "env": {
+    "ELASTICSEARCH_URL": "http://elasticsearch-9200.service.consul:9200",
```
tayzlor: this probably needs to use ```{{ consul_domain }}``` just in case it gets changed service.consul wont work

roles/dcos_cli/vars/logstash.yml
```diff
@@ -0,0 +1,5 @@
+dcos_cli_app_logstash_enabled: "{{ elk_enabled }}"
+dcos_cli_app_logstash_mem: 512
+dcos_cli_app_logstash_cpus: 0.5
+dcos_cli_app_logstash_instances: 1
+dcos_cli_app_logstash_elasticsearch: elasticsearch-9200.service.consul:9200
```
tayzlor: should use ```consul_domain```

roles/mesos/defaults/main.yml
```diff
@@ -2,7 +2,7 @@
 # Common
 consul_dir: /etc/consul.d
 mesos_executor_registration_timeout: 10mins
-mesos_cluster_name: "Cluster01"
+mesos_cluster_name: "{{ mesos_cluster_name }}"
```
tayzlor: seems a bit odd to change this here

tayzlor:
How do we handle log files that are on the host by default? Thining (there may be others)...

- consul
- docker
- weave?

roles/mesos/defaults/main.yml
wallies: i thought it was odd we were setting this explicitly here when we have the global var

roles/dcos_cli/defaults/main.yml
```diff
@@ -10,7 +10,19 @@ dcos_cli_frameworks_list:
   - chronos
 dcos_cli_apps_list:
   - prometheus
+  - elasticsearch
+  - logstash
+  - kibana
 
 # apps
 prometheus_image: prom/prometheus
```
enxebre: * Why are we breaking the name convention here?
* (Also for prometheus) Why dont we put the app specific variables under vars/app-name.yml as we are 
```
- include_vars: "{{ item }}.yml"
  with_items:
    - "{{ dcos_cli_apps_list }}"
```

roles/dcos_cli/templates/elasticsearch.json.j2
```diff
@@ -0,0 +1,45 @@
+{
+  "id": "elasticsearch",
+  "instances": {{ dcos_cli_app_elasticsearch_instances }},
+  "cpus": {{ dcos_cli_app_elasticsearch_cpus }},
+  "mem": {{ dcos_cli_app_elasticsearch_mem }},
+  "labels": {
+    "DCOS_PACKAGE_NAME": "elasticsearch",
+    "DCOS_PACKAGE_IS_FRAMEWORK": "true"
```
enxebre: is it true?

roles/dcos_cli/templates/kibana.json.j2
```diff
@@ -0,0 +1,39 @@
...
+        "protocol": "tcp" 
+      }]
+    }
+  },
+  "instances": {{ dcos_cli_app_kibana_instances }},
+  "cpus": {{ dcos_cli_app_kibana_cpus }},
+  "mem": {{ dcos_cli_app_kibana_mem }},
+  "labels": {
+    "DCOS_PACKAGE_NAME": "kibana",
+    "DCOS_PACKAGE_IS_FRAMEWORK": "true"
```
enxebre: is it true?

roles/dcos_cli/templates/logstash.json.j2
```diff
@@ -0,0 +1,46 @@
+{
+  "id": "logstash",
+  "cmd": "/bin/bash -c \"/opt/logstash/bin/plugin install logstash-input-salesforce; logstash -f /config-dir/logstash.conf\"",
+  "instances": {{ dcos_cli_app_logstash_instances }},
+  "cpus": {{ dcos_cli_app_logstash_cpus }},
+  "mem": {{ dcos_cli_app_logstash_mem }},
+  "labels": {
+    "DCOS_PACKAGE_NAME": "logstash",
+    "DCOS_PACKAGE_IS_FRAMEWORK": "true"
```
enxebre: is it true?

roles/logstash/tasks/main.yml
```diff
@@ -0,0 +1,83 @@
...
+    state: started
+    restart_policy: "{{ logspout_restart_policy }}"
+    net: "{{ logspout_net }}"
+    ports:
+    - "8000:8000"
+    env:
+      ROUTE_URIS: logstash://{{ logstash_syslog_host }}:5000
+      LOGSPOUT: ignore
+    volumes:
+    - "/var/lib/docker/:/var/run/docker.sock:ro"
```
enxebre: If we want this to join the weave net and for consitency we could set this at the variables level like:
``` logsoupt_docker_host: {{ docker_host }} ```

**Comment by wallies** [comment]
yeh it is because elasticsearch exposes multiple ports. I'm not sure what i will do about elasticsearch yet, i want to add an encrypted volume, so we can achieve data security at rest. i have just had a go at moving to logstash-forwarder instead of logspout and that will work for container logs and host logs, to achieve data security in transport between hosts and logstash.

File: roles/dcos_cli/templates/logstash.json.j2 (hunk 7 above)
Comment (wallies) [inline roles/dcos_cli/templates/logstash.json.j2]: well even though its not using a dcos framework, for purposes of labelling its a framework for the platform, and not an app.

File: roles/dcos_cli/defaults/main.yml (hunk 4 above)
Comment (wallies) [inline roles/dcos_cli/defaults/main.yml]: i agree with you here, will change.

**Comment by enxebre** [comment]
would be nice if we can add some tests for this

**Comment by pmbauer** [comment]
I question this approach, namely, spinning up elasticsearch in a docker container via marathon backed with ephemeral storage on the slaves.
We are currently spinning up our own elasticsearch on a separate server in the Apollo vpc backed by EBS with plans to move to AWS hosted elasticsearch in the near future.

**Comment by wallies** [comment]
i plan to create a encrypted EBS volume and attach it to the elasticsearch docker containers, then manage this with flocker. I do like the idea of AWS hosted elasticsearch though.
@enxebre tests on there way

File: roles/consul/templates/consul.json.j2 [hunk 9]
```diff
@@ -14,5 +14,9 @@
   "bootstrap_expect": {{ consul_bootstrap_expect }},
 {% endif %}
   "data_dir": "/var/lib/consul",
-  "ui_dir": "/opt/consul-ui"
+  "ui_dir": "/opt/consul-ui",
+  "recursors": [ {{ consul_retry_join }} ],
```
enxebre: When "consul_syslog" not enabled this is going to break due to the comma here ```"recursors": [ {{ consul_retry_join }} ],```yielding an invalid json.

roles/logstash/handlers/main.yml
```diff
@@ -0,0 +1,7 @@
+---
+# handlers file for docker
+- name: restart zookeeper
```
tayzlor: restart zookeeper? should this be logstash?

wallies: nope, needs to be zookeeper, as logstash isnt up when zookeeper starts so times out trying to send logs to logstash. Going to change it to use stdout/stderr so fits with the rest of the apps

tayzlor: seems like this should use the global 'handlers' role then and the code should move there, as it confused me being here

wallies:
closing, will redo PR based on new devel branch work based on coreos.

PR #553, opened 2015-12-04
Upgraded to marathon 0.13.0 and weave 1.3.1

ravbaba:


enxebre:
nice, like you just decide to go for twenty :D should we just remove the atlas_artifact_version variables as it's not been used anymore?

PR #559, opened 2015-12-07
Datadog.

enxebre:
We should be able to combine this with https://github.com/tsenart/vegeta and get some feeback with fancy populated dashboards


tayzlor:
code lgtm

wallies:
lgtm

tayzlor:
:ship:

PR #584, opened 2016-01-04
change to fix marathon spec test

wallies:


tests/spec/marathon/marathon_spec.rb
```diff
@@ -24,6 +24,9 @@
 
 describe command("curl -s -XPOST 127.0.0.1:8080/v2/apps -d@spec/marathon/sample.json -H \"Content-Type: application/json\" && sleep 10") do
   its(:exit_status) { should eq 0 }
+  it {
```
enxebre: should we just increase the sleep 10 above, or just remove it?

wallies: didnt see this. my way works but looks hacky. ill just increase the sleep above this.

PR #591, opened 2016-01-05
Fix #577 - set min version for channels

### LINKED ISSUE #577: Move config.vm.box_version to be specific to versions based on which coreOS channel we are using
This variable should be pulled out of vagrant.yml file and will be different depending on what we set the coreOS channel to be (stable/beta/alpha)

tayzlor:


enxebre:
lgtm

Issue #580, opened 2016-01-04
Update consul service + health checks for coreOS and consul 0.6


tayzlor:
With the upgrade of consul to 0.6 we have better health check support for docker via docker + tcp checks - https://hashicorp.com/blog/consul-0-6.html#health-checks

We should be able to migrate some checks to this and we also need to ensure our checks still work with coreOS for the platform components we have.

PR #598, opened 2016-01-05
Initial commit for #580 - add consul health checks based on consul 0.6

tayzlor:
this could either go in as in or we also need to add - 

- [ ] mesos slave
- [ ] weave 
- [ ] consul agent (do we need this?)


tayzlor:
merging as is will add the remaining checks to #580

PR #599, opened 2016-01-05
Fix #575 - remove rebuild_container in favour of just using systemd 

### LINKED ISSUE #575: Remove rebuild_container tasks and variables in favour of just using systemd
For example looking at https://github.com/Capgemini/Apollo/blob/coreos/roles/mesos/templates/mesos-master.service.j2#L7-L25 it will pull the correct image and start it correctly on a service restart (if the template file is updated with a new docker image)

tayzlor:


roles/marathon/defaults/main.yml
```diff
@@ -12,7 +12,6 @@ marathon_java_settings: '-Xmx512m -Xms512m -XX:+HeapDumpOnOutOfMemoryError'
 marathon_artifact_store: 'file:///store'
 marathon_artifact_store_dir: '/etc/marathon/store'
 marathon_server_zk_group: marathon_servers
-marathon_rebuild_container: False
```
enxebre: we loose ability to recreate containers with this, should we create ```ensure marathon is restarted``` task and ``` ```marathon_restart_service: False``` ?

wallies: systemd takes over this with the ExecPre

tayzlor: should be recreateable by changing the image name (which will update the systemd file) and reload the service with the new image in place.

Issue #594, opened 2016-01-05
update terraform aws plans to be compatible with coreos


tayzlor:

tayzlor:
Closing via https://github.com/Capgemini/Apollo/pull/603

PR #601, opened 2016-01-06
#594 - add terraform plans for coreOS on AWS public cloud

tayzlor:


tayzlor:
Just noticed we need to handle the EBS volume. 

That might need to be handled slightly differently. See https://coreos.com/os/docs/latest/mounting-storage.html

https://coreos.com/os/docs/latest/booting-on-ec2.html

bootstrap/common.sh
```diff
@@ -131,13 +131,14 @@ ansible_playbook_run() {
...
     ${ANSIBLE_LOG} --extra-vars "$( get_apollo_variables  APOLLO_)" \
     ${ANSIBLE_EXARGS:-} \
-    --sudo site.yml
+    site.yml
   popd
 }
 
 install_contributed_roles() {
   pushd "${APOLLO_ROOT}"
     ansible-galaxy install --force -r contrib-plugins/plugins.yml
```
enxebre: wonder if should unify this two

wallies: yeh definitely

PR #607, opened 2016-01-08
docker service override.

enxebre:


roles/docker/templates/docker.service.j2
```diff
@@ -0,0 +1,15 @@
+[Unit]
+Description=Docker Application Container Engine
+Documentation=http://docs.docker.com
+After=docker.socket early-docker.target network.target
+Requires=docker.socket early-docker.target
+
+[Service]
+EnvironmentFile=-/run/flannel_docker_opts.env
```
tayzlor: flannel?

roles/docker/tasks/main.yml
```diff
@@ -1,35 +1,17 @@
...
-#- name: configure docker temporary directory
-#  sudo: yes
-#  lineinfile:
-#    dest: /etc/sysconfig/docker
-#    state: present
-#    line: 'DOCKER_TMPDIR=\"{{ docker_tmp_dir }}\"'
-#  notify:
-#    - restart docker
-
-#- name: configure docker proxy
```
tayzlor: where do we handle these options now ?

roles/docker/tasks/main.yml
```diff
@@ -1,35 +1,17 @@
 ---
 # tasks file for docker
-- name: configure docker consul dns
+- name: deploy docker service
   sudo: yes
-  lineinfile:
-    dest: /etc/sysconfig/docker
-    regexp: ^OPTIONS=
-    line: 'OPTIONS=\"--storage-driver=overlay --graph={{ docker_graph_dir }} --dns 172.17.0.1 --dns 8.8.8.8 --dns-search service.{{ consul_domain }} \"'
```
tayzlor: where are these options getting handled now?

roles/docker/templates/docker.service.j2
enxebre: This is just a copy of /usr/lib/systemd/system/docker.service as suggested in here https://coreos.com/os/docs/877.1.0/customizing-docker.html we can inject drop it settings just by adding files in ``` /etc/systemd/system/docker.service.d/ ```

roles/docker/tasks/main.yml
enxebre: this file is ignored by coreOS. This is handle now in  roles/docker/templates/docker.service.j2
We could override DOCKER_OPTS environment variable here /etc/systemd/system/docker.service.d/
but coreOS is giving a memory error message when using more than one flag there
https://coreos.com/os/docs/877.1.0/customizing-docker.html

PR #610, opened 2016-01-08
Registrator container

enxebre:
Moving this into a systemd service


roles/registrator/tasks/main.yml
```diff
@@ -0,0 +1,28 @@
+---
+
+- name: wait for weave socket to be ready.
```
tayzlor: this should be in the weave role (I think it is already)

roles/registrator/tasks/main.yml
```diff
@@ -0,0 +1,28 @@
...
+  wait_for:
+    port: 6783
+    delay: 10
+
+- name: destroy old marathon container
```
tayzlor: if we have a systemd init file then we dont need either of these tasks - just deploy the init file and start it. no need for rebuild_container stuff

PR #616, opened 2016-01-11
Add check script for marathon DNS for traefik

### LINKED ISSUE #595: Traefik container does not grab marathon config if consul DNS isn't up properly
Occassionally the traefik container starts up while the consul container is still booting and tries to access marathon.service.consul resulting in - 

```
time="2016-01-05T13:43:47Z" level=error msg="Failed to register for subscriptions, all the Marathon hosts are presently down"
time="2016-01-05T13:43:47Z" level=error msg="Failed to create a client for marathon, error: all the Marathon hosts are presently down"
```

We need to add a wait / check in around consul DNS

tayzlor:
Fixes #595

### DIFF
--- a/roles/consul/tasks/main.yml
+++ b/roles/consul/tasks/main.yml
@@ -12,6 +12,11 @@
   tags:
     - consul
 
+- name: wait for consul to listen
+  wait_for:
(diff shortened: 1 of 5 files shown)


enxebre:
code looks good to me

PR #622, opened 2016-01-14
Elk docker

wallies:
Gets elb logs into s3 and inputs them into logstash as a bonus


roles/datadog/README.md
```diff
@@ -0,0 +1,81 @@
+Ansible Datadog Role
```
tayzlor: is this role meant to be committed?

PR #623, opened 2016-01-20
Fix datadog integration

wallies:


contrib-plugins/playbook.yml
```diff
@@ -1,10 +1,14 @@
 # Sample plugins playbook.
 - hosts: all:!role=bastion
+
   tasks:
-    # https://github.com/DataDog/dd-agent/blob/master/conf.d/docker.yaml.example
```
enxebre: don we stil need this?

terraform/aws/public-cloud/etcd_discovery_url.txt
```diff
@@ -1 +1 @@
-https://discovery.etcd.io/fafb16ac572cae5895c6523e7b3bd640
\ No newline at end of file
+https://discovery.etcd.io/e141564df4f4f85905a0d9cb447c6c2d
```
enxebre: I guess we need to git update-index --assume-unchanged this file

enxebre:
Otherwise than that looks good to me @wallies

contrib-plugins/playbook.yml
wallies: its in the ansible role, as we cant run pre_tasks until 2.0

PR #626, opened 2016-01-29
#374 Added timeouts to consul health checks

drpauldixon:


wallies:
@drpauldixon can you make this PR to devel branch instead. we have moved all work to devel for now

PR #635, opened 2016-02-19
Docker daemon extra variables.

enxebre:


roles/docker/templates/docker.service.j2
```diff
@@ -8,13 +8,14 @@ Requires=docker.socket early-docker.target
...
 Environment="HTTP_PROXY={{ http_proxy }}"
+Environment="NO_PROXY={{ docker_proxy_exceptions }}"
 {% endif %}
 Environment="DOCKER_TMPDIR={{ docker_tmp_dir }}"
 
 MountFlags=slave
 LimitNOFILE=1048576
 LimitNPROC=1048576
-ExecStart=/usr/lib/coreos/dockerd daemon {{ docker_storage_config }} {{ docker_dns_config }} --host=fd:// $DOCKER_OPTS $DOCKER_OPT_BIP $DOCKER_OPT_MTU $DOCKER_OPT_IPMASQ
+ExecStart=/usr/lib/coreos/dockerd daemon {{ docker_endpoints }} {{ docker_bip}} {{ docker_storage_config }} {{ docker_dns_config }} {{ docker_registry }} --host=fd:// $DOCKER_OPTS $DOCKER_OPT_BIP $DOCKER_OPT_MTU $DOCKER_OPT_IPMASQ
```
tayzlor: would probably call the variable explicitly ```docker_bridge_ip``` so its obvious what it is - also v minor but a space after the p and before the }

Issue #651, opened 2016-02-26
Use public networking for digitalocean


sheerun:
"Private networking" on digital ocean isn't private at all:
http://incoherency.co.uk/blog/stories/digital-ocean-private-network.html

So we might as well use public networking and allow masters to span across datacenters, forexample 3 masters in nyc1, nyc2, and nyc3. Currently you can't do it because private shared networking cannot reach across datacenters even in the same region.

tayzlor:
@sheerun agreed. thanks for the heads up - we'll push a fix into devel for that shortly

tayzlor:
Closing via #654

PR #654, opened 2016-02-28
Use public cloud for digital ocean and secure with TLS

sheerun:
"Private network" on digital ocean is really a shared private network.

It means all other hosts in datacenter, even of other users, can access deployed nodes on digitalocean. It means deploying on public ips is not less secure than deploying on private ones.

In this setup we additionaly secure etcd with wildcard tls certs.

As an additional bonus, we can not deploy servers across datacenters.

Also, we generate initial ssh key for digital ocean dynamically, for better security, and export it to "id_rsa" file for easy ssh access.


sheerun:
This closes #651

terraform/digitalocean/main.tf
```diff
@@ -13,11 +12,82 @@ provider "digitalocean" {
...
+  algorithm = "RSA"
+}
+
+resource "tls_self_signed_cert" "ca" {
+  key_algorithm = "RSA"
+  private_key_pem = "${tls_private_key.ca.private_key_pem}"
+
+  subject {
+    common_name = "*"
+    organization = "skybase"
```
tayzlor: minor: wonder if we should make this  ```${var.org_name}``` and default it to "apollo" or similar.

sheerun:
Made the fix you suggested :)

tayzlor:
Thanks!

PR #656, opened 2016-02-28
Rackspace

bkarypid:
This is an Openstack implementation for Apollo, geared towards and tested in Rackspace public cloud.


inventory/inventory
```diff
@@ -17,7 +17,8 @@ mesos_slaves
 load_balancers
 
 [all:vars]
-ansible_python_interpreter="PATH=/home/core/bin:$PATH python"
+# The only user able to ssh to a newly created Rackspace instance is root (at least initially), so we need to update the python PATH to accommodate for that.
+ansible_python_interpreter="PATH=/home/core/bin:/root/bin:$PATH python"
```
wallies: is this the case even for coreos machine

wallies:
have you looked at the other openstack branch we had started? also can we rename this to rackspace instead of openstack, as there are little differences between rackspace and openstack.

bkarypid:
I did have a look at the other branch, but it was quite outdated and I thought it would be easier to just start from scratch really, plus as you said there are some differences in the Rackspace implementation of Openstack, especially in networking, which Terraform can't handle for Rackspace at the moment. Details on these limitations can be found in the getting started doc. I'll change the directories/names from openstack to rackspace as well and push back(i had them like this initially).

bkarypid:
Done and retested in Rackspace, I can give you a demo tomorrow before you consider merging if you'd like.

wallies:
yep demo would be good. two questions, the default user, why are we using root, also can you look how we mounted storage for each slave for /var/lib/docker in aws

inventory/inventory
bkarypid: Yes it is unfortunately, regardless of the image type / OS used (eg. ubuntu,centos,coreos etc.), in Rackspace the first time you try to ssh in one of the machines you can only do so as root. You can change that behaviour from within the box after you ssh as root initially of course. See here for more info: https://support.rackspace.com/how-to/basic-cloud-server-security/

bkarypid:
See my comment about the root user for Rackspace above. I already had a look at the additional cloud init config for mesos slaves in aws, however I'm not sure whether the same thing can be done for Rackspace, at least the ebs formatting bit, so perhaps the mounting process might need to be a bit different. We can discuss more about this tomorrow.

wallies:
so looks like terraform openstack has support for volumes on instances like this 
https://github.com/CiscoCloud/mantl/blob/master/terraform/openstack/instance/main.tf, you can then reference the volume device in the cloud config

sheerun:
shouldn't it be named "openstack" instead of "rackspace"? Isn't the only variable for terraform script the identity url?

wallies:
it could be but rackspace has some little differences like you cant create networks, it uses the identity url, and authentication to this differently to other openstack providers

bootstrap/common.sh
```diff
@@ -31,7 +31,7 @@ verify_prereqs() {
 check_terraform_version() {
   local IFS='.'
   local current_version_string="${2:-$( terraform --version | awk 'NR==1 {print $2}' )}"
-  local requirement_version_string=${1:-0.6.12}
+  local requirement_version_string=${1:-0.6.10} # This is to support Rackspace. Access IPs are not properly exported by the Terraform Openstack provider in versions > 0.6.10. See https://github.com/hashicorp/terraform/issues/5358
```
tayzlor: this won't stop people from installing 6.12 and hitting this issue, since it only checks the lower version and does not check the upper (e.g. that you don't have anything installed over 6.10)

bootstrap/rackspace/config-default.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+export TF_VAR_region=${TF_VAR_region:-LON}
+export TF_VAR_auth_url=${TF_VAR_auth_url:-https://identity.api.rackspacecloud.com/v2.0}
+export TF_VAR_mesos_slaves=${TF_VAR_mesos_slaves:-1}
```
tayzlor: dont think this var is needed (should be set to a default in the terraform config)

bootstrap/rackspace/config-default.sh
```diff
@@ -0,0 +1,13 @@
...
+
+# Overrides default folder in Terraform.py inventory.
+export TF_VAR_STATE_ROOT="${APOLLO_ROOT}/terraform/${APOLLO_PROVIDER}"
+
+export APOLLO_consul_dc=${APOLLO_consul_dc:-$TF_VAR_region}
+export APOLLO_mesos_cluster_name=${APOLLO_mesos_cluster_name:-$TF_VAR_region}
+export APOLLO_ansible_ssh_user=root
+export APOLLO_traefik_network_interface=eth0
```
tayzlor: again this should be the default value so no need to export it explicitly here

inventory/inventory
tayzlor: can we set a var group that is explicit to "rackspace" somehow, since there will be nothing in /root/bin elsewhere

bootstrap/rackspace/config-default.sh
bkarypid: It's already the default in the terraform config, so you're right it's unnecessary here (all of them have defaults tbh, since this provider is geared towards Rackspace). Will remove.

bootstrap/common.sh
bkarypid: Agreed, just wanted to be able to test it. Thought about adding a Rackspace specific terraform version check function (0.6.8-0.6.10 are acceptable atm), but given that this bug has been identified, fixed and will be released in 0.6.13, I wouldn't worry too much about it. When 0.6.13 is released we can probably use that as a requriement. See https://github.com/hashicorp/terraform/issues/5358

tayzlor: ah if it's in 0.6.13 then fine not too worried and we can bump it then.

bkarypid: Changed it back to 0.6.12. I'm using the Openstack provider built from source with the fix implemented though, so I would wait for TF 0.6.13 before merging.

bkarypid:
Sorted out the ssh user issue via CoreOS cloud config for both mesos masters and slaves, now the only user allowed to ssh to the CoreOS instances in Rackspace is core, so that's what Ansible uses as well (and the PATH to python/pypy is the same as in the other providers). Also, thankfully the Terraform Openstack provider's openstack_blockstorage_volume_v1 resource works fine with Rackspace, so storage for /var/lib/docker is also now mounted in the mesos slave instances.

bootstrap/rackspace/config-default.sh
```diff
@@ -0,0 +1,15 @@
...
+
+# Overrides default folder in Terraform.py inventory.
+export TF_VAR_STATE_ROOT="${APOLLO_ROOT}/terraform/${APOLLO_PROVIDER}"
+
+export TF_VAR_region=${TF_VAR_region:-LON}
+
+export APOLLO_consul_dc=${APOLLO_consul_dc:-$TF_VAR_region}
+export APOLLO_mesos_cluster_name=${APOLLO_mesos_cluster_name:-$TF_VAR_region}
+export APOLLO_ansible_ssh_user=core
+export APOLLO_traefik_network_interface=eth0
```
enxebre: just notice you dont really need this one as it's default, does not hurt to have it though

terraform/rackspace/modules/keypair/main.tf
```diff
@@ -0,0 +1,14 @@
...
+# SSH keypair resource
+resource "openstack_compute_keypair_v2" "default" {
+  name       = "${var.key_name}"
+  public_key = "${file(var.public_key_file)}"
+}
+
+# Outputs
+output "keypair_name" {
+  value = "${openstack_compute_keypair_v2.default.name}"
+}
```
enxebre: If using sublime you can auto include end of file line by setting ```ensure_newline_at_eof_on_save```

terraform/rackspace/modules/mesos_masters/main.tf
```diff
@@ -0,0 +1,60 @@
...
+  name              = "apollo-mesos-master-${count.index}"
+  flavor_id         = "${var.instance_type}"
+  image_id          = "${var.image_id}"
+  count             = "${var.masters}"
+  key_pair          = "${var.key_pair}"
+  network           =
+    {
+      uuid          = "${var.public_network_id}"
+      name          = "${var.public_network_name}"
+      # access_network = "true"
```
enxebre: Can we remove this commented bits?

terraform/rackspace/modules/mesos_slaves/main.tf
```diff
@@ -0,0 +1,79 @@
...
+  name              = "apollo-mesos-slave-${count.index}"
+  flavor_id         = "${var.instance_type}"
+  image_id          = "${var.image_id}"
+  count             = "${var.slaves}"
+  key_pair          = "${var.key_pair}"
+  network           =
+    {
+      uuid           = "${var.public_network_id}"
+      name           = "${var.public_network_name}"
+      # access_network = "true"
```
enxebre: Can we remove commented bits?

enxebre:
Just dropped few really minor comments, looks cool to me, could we just clean and squash the git history little bit?

bootstrap/rackspace/config-default.sh
```diff
@@ -0,0 +1,15 @@
...
+export TF_VAR_password=${TF_VAR_password:?"Need to set TF_VAR_password non-empty"}
+
+# Overrides default folder in Terraform.py inventory.
+export TF_VAR_STATE_ROOT="${APOLLO_ROOT}/terraform/${APOLLO_PROVIDER}"
+
+export TF_VAR_region=${TF_VAR_region:-LON}
+
+export APOLLO_consul_dc=${APOLLO_consul_dc:-$TF_VAR_region}
+export APOLLO_mesos_cluster_name=${APOLLO_mesos_cluster_name:-$TF_VAR_region}
+export APOLLO_ansible_ssh_user=core
```
tayzlor: shouldn't need this as it should be default?

docs/getting-started-guides/rackspace.md
```diff
@@ -0,0 +1,104 @@
...
+
+#### Set config
+
+Configuration can be set via environment variables.
+
+All variables following the pattern "TF_VAR_" will be available for Apollo in terraform, see [https://github.com/hashicorp/terraform/pull/1621#issuecomment-100825568](https://github.com/hashicorp/terraform/pull/1621#issuecomment-100825568)
+
+All variables following pattern "APOLLO_" will be available for Apollo in ansible.
+
+Certain default config options for Rackspace can be found in `bootstrap/openstack/config-default.sh`
```
tayzlor: should be ```bootstrap/rackspace``` according to the file above

terraform/rackspace/modules/etcd_discovery/main.tf
```diff
@@ -0,0 +1,23 @@
+# Variables
```
tayzlor: seems like this can be moved to a more top-level which cuts across all providers (and we can re-use in other providers) - thoughts? can we create a follow-up ticket to address this?

terraform/rackspace/modules/mesos_slaves/slave-cloud-config.yml.tpl
```diff
@@ -0,0 +1,64 @@
...
+    # listen on both the official ports and the legacy ports
+    # legacy ports can be omitted if your application doesn't depend on them
+    listen-client-urls: "http://0.0.0.0:2379,http://0.0.0.0:4001"
+    listen-peer-urls: "http://$private_ipv4:2380,http://$private_ipv4:7001"
+    # Discovery is populated by Terraform
+    discovery: ${etcd_discovery_url}
+  units:
+    - name: etcd2.service
+      command: start
+    - name: fleet.service
```
tayzlor: lets remove fleet from here since we don't use it

tayzlor:
dropped a few minor comments, code looks mostly good, can you fix and squash?

One last question from me - what's the difference here between 'rackspace' and a 'pure openstack' approach? How easy is it to create a pure openstack approach from this? could it be made modular? can we address that in a follow-up issue?

bootstrap/rackspace/config-default.sh
bkarypid: Tested without it and Ansible couldn't connect via ssh, so had to re-add it and it worked. I can also see that it is present in the aws and digitalocean providers as well, so it's probably necessary at the moment.

terraform/rackspace/modules/etcd_discovery/main.tf
bkarypid: This can be reused, as the null resource isn't tied to any provider. A few minor changes would need to happen in the user data resources in the other providers, but nothing big. My only initial concern with this approach was the fact that the file contents were being passed to the user data template as a var, which would naturally destroy all instances when changing counts, but seeing your issue/comment about the possibility of using a production cluster setup makes this approach acceptable I think.

bkarypid:
Closing this to open another one to master. I believe that the modules could be reused for a pure Openstack implementation, in a similar fashion to aws public / private setup. The main difference would be that an Openstack implementation can use networking resources, which isn't possible with Rackspace. The modules should be able to be reused by both providers, which would have different main.tf files, for example the RS provider would pass the (already existing) network IDs to the compute resource modules from variables, whereas the Openstack provider would pass them by reading outputs from eg. a networking module. A slight difference (according to the docs) is that when specifying a network for a compute instance resource, either the uuid (network id) or the network name can be passed (one would suffice), however Rackspace needs both. Both these attributes are exported by the networking/subnet resources though, so we could still use the same setup for both, at least in theory. It might be a bit redundant, but would allow for reusability.

PR #660, opened 2016-03-01
Use generated id_rsa file for digitalocean ssh access

sheerun:
This fixes ansible connection with digitalocean


wallies:
lgtm. cheers

PR #669, opened 2016-03-08
Initial work on moving bastion instance to coreOS host

tayzlor:
Still todo - 

- [x] proper testing of ansible 2.0 (dependency for this PR since ```delegate_to``` is broken (see https://github.com/ansible/ansible/issues/12772) with sending variables in 1.9.4
- [x] move coreOS bootstrapping role down into terraform (partly done with the bootstrap.sh shell script but could be done across all the other instances)


tayzlor:
Think we should move the coreOS bootstrapping down for the other clouds (and remove the ansible module) in a follow-up PR

PR #672, opened 2016-03-09
Upgrade traefik container default image

tayzlor:


wallies:
can we move to this image instead containous/traefik

wallies:
nice +1

PR #673, opened 2016-03-09
Replace registrator with direct bridge to consul via mesos_consul

tayzlor:


wallies:
if this works, will be awesome to get rid of registrator

tayzlor:
it works of course :hankey: any other feedback/comments or lgtm?

wallies:
Lgtm +1

roles/mesos_consul/tasks/main.yml
```diff
@@ -0,0 +1,23 @@
+---
+
+- name: deploy mesos_consul service
+  become: yes
```
enxebre: Do we need both become?

roles/mesos_consul/templates/mesos_consul.service.j2
```diff
@@ -0,0 +1,22 @@
+[Unit]
+Description=Mesos Consul
```
enxebre: Thinking if we should run this through the weave proxy, probably does not matter

enxebre: Agree we @wallies if it works lgtm 🐗🐗

PR #680, opened 2016-03-22
Add elk to apollo

wallies:


joe1chen:
I was playing around with this branch and notices a few issues that prevented me from getting it working:

1) Hard-coded references to ```eu-west-1``` in ```roles/logstash/templates/logstash.conf.j2```
2) Hard-coded s3 bucket (```apollo-elb-logs ```) in ```terraform/aws/elb/main.tf```. Please correct me if I'm wrong but doesn't the s3 policy need to be parameterized so that it can be used in my aws account?

wallies:
@joe1chen yes you are correct. Have fixed the hardcoded referenced region and bucket name. That s3 policy should work for the elb to log to s3 in your account. cheers

joe1chen:
@wallies I still see the hardcoded references in this pull request.  Am I missing something?

PR #682, opened 2016-03-23
Proper fix for the apollo-down bug.

joe1chen:
The previous bug fix was not properly checked in.  This should fix the output of ```terraform destroy``` from being redirected to the ```etcd_discovery_url.txt``` file.

### DIFF
--- a/bootstrap/aws/private-cloud/util.sh
+++ b/bootstrap/aws/private-cloud/util.sh
@@ -51,9 +51,7 @@ apollo_down() {
   pushd "${APOLLO_ROOT}/terraform/${APOLLO_PROVIDER}"
     terraform destroy -var "access_key=${TF_VAR_access_key}" \
       -var "key_file=${TF_VAR_key_file}" \
-      -var "region=${TF_VAR_region}" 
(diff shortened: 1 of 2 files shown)


wallies:
your last fix is actually what we intended to do and fixes the issue. so will reject this one

joe1chen:
@wallies Can you explain why the last fix was actually what was intended?  With the first fix, then output of terraform destroy is output to ```etcd_discovery_url.txt```.

Contents of ```etcd_discovery_url.txt```:
```
Do you really want to destroy?
  Terraform will delete all your managed infrastructure.
  There is no undo. Only 'yes' will be accepted to confirm.

  Enter a value: 
```

On the command-line, I end up seeing no output at all.  It seems to hang because it is waiting for the user to type ```yes``` but without any output in the console user has no idea that they have to type ```yes```.

wallies:
@joe1chen when you destroy the cluster it should send an empty string to the etcd_discovery_url.txt. I have just tested it with your last PR #681 and it has the intended behaviour.

joe1chen:
@wallies Ok, hmm let me re-test this again, I'll see if I can figure out what's going on in my environment.

Issue #678, opened 2016-03-21
Unable to bootstrap on aws


joe1chen:
I have set ```APOLLO_PROVIDER=aws``` and all the other environment variables, but when I run ```/bin/bash bootstrap/apollo-launch.sh``` it results in an error.

Output:
```
/bin/bash bootstrap/apollo-launch.sh
bootstrap/apollo-launch.sh: line 10: bootstrap/../bootstrap/../bootstrap/aws/util.sh: No such file or directory
```

Looks like the aws files got moved around?  There is now a ```aws/public-cloud/util.sh``` and ```aws/private-cloud/util.sh```.  How can I fix the error?

wallies:
@joe1chen if you set APOLLO_PROVIDER=aws/public-cloud or APOLLO_PROVIDER=aws/private-cloud it should work

joe1chen:
Thanks, the docs need to be updated.

tayzlor:
@joe1chen yes - they do - they're a bit out of date since we merged some PRs to move to CoreOS. Please feel free to send in a PR to update the docs if you can!

joe1chen:
@tayzlor Will definitely send a PR once I get everything working.  I'm now blocked on issue #679.

PR #683, opened 2016-03-23
Update aws documentation

joe1chen:
One thing to note is that there are references to both ```${var.public_key_file}``` (in ```main.tf```) and as well as ```TF_VAR_key_file``` (in ```util.sh```), so both ```TF_VAR_public_key_file``` and ```TF_VAR_key_file``` need to be defined.

It might be better to merge these variables into a single one.


joe1chen:
BTW, this fixes #678.

wallies:
cheers @joe1chen

Issue #679, opened 2016-03-21
Error during apollo launch on AWS


joe1chen:
I'm using the ```aws/public-cloud``` startup scripts.  On the last step, I'm seeing this error:

```

TASK: [weave | deploy weave service] ****************************************** 
fatal: [apollo-mesos-slave-2] => {'msg': "AnsibleUndefinedVariable: One or more undefined variables: 'dict object' has no attribute 'ansible_default_ipv4'", 'failed': True}
fatal: [apollo-mesos-slave-3] => {'msg': "AnsibleUndefinedVariable: One or more undefined variables: 'dict object' has no attribute 'ansible_default_ipv4'", 'failed': True}
fatal: [apollo-mesos-slave-1] => {'msg': "AnsibleUndefinedVariable: One or more undefined variables: 'dict object' has no attribute 'ansible_default_ipv4'", 'failed': True}
fatal: [apollo-mesos-slave-2] => {'msg': 'One or more items failed.', 'failed': True, 'changed': False, 'results': [{'msg': "AnsibleUndefinedVariable: One or more undefined variables: 'dict object' has no attribute 'ansible_default_ipv4'", 'failed': True}]}
fatal: [apollo-mesos-slave-3] => {'msg': 'One or more items failed.', 'failed': True, 'changed': False, 'results': [{'msg': "AnsibleUndefinedVariable: One or more undefined variables: 'dict object' has no attribute 'ansible_default_ipv4'", 'failed': True}]}
fatal: [apollo-mesos-slave-1] => {'msg': 'One or more items failed.', 'failed': True, 'changed': False, 'results': [{'msg': "AnsibleUndefinedVariable: One or more undefined variables: 'dict object' has no attribute 'ansible_default_ipv4'", 'failed': True}]}
fatal: [apollo-mesos-master-1] => {'msg': "AnsibleUndefinedVariable: One or more undefined variables: 'dict object' has no attribute 'ansible_default_ipv4'", 'failed': True}
fatal: [apollo-mesos-master-1] => {'msg': 'One or more items failed.', 'failed': True, 'changed': False, 'results': [{'msg': "AnsibleUndefinedVariable: One or more undefined variables: 'dict object' has no attribute 'ansible_default_ipv4'", 'failed': True}]}
fatal: [apollo-mesos-slave-0] => {'msg': "AnsibleUndefinedVariable: One or more undefined variables: 'dict object' has no attribute 'ansible_default_ipv4'", 'failed': True}
[...]

PR #685, opened 2016-03-28
Fixes ansible error accessing "ansible_default_ipv4" on Mac OS X

joe1chen:
The underlying bug is a problem with ansible (https://github.com/ansible/ansible/issues/9260).  This bug is responsible for https://github.com/Capgemini/Apollo/issues/679.

The workaround for this issue is to simply add:

```
- hosts: all
  tasks: []
```

to ```site.xml``` before we try to access ```ansible_default_ipv4```.

Without this workaround, running the bootstrap script on OS X systems seems to consistently fail.

### DIFF
--- a/site.yml
+++ b/site.yml
@@ -20,6 +20,9 @@
 
 - include: playbooks/coreos-bootstrap.yml
 
+- hosts: all
+  tasks: []
+
 - hosts: all:!role=bastion
   roles:
     - docker


joe1chen:
I'm still getting the ansible_default_ipv4 errors occasionally with this fix, so for now I'm going to close this PR.

PR #689, opened 2016-04-05
Support single availability zone on aws

joe1chen:
Allow subnets to be set from environment variables.  This is required to support running in a single availability zone. 

Currently if you try to run in a single az, you get an error with the ELB

Environment variables:
```
export TF_VAR_availability_zones=us-east-1a
```

Error:
```
aws_elb.elb: InvalidConfigurationRequest: ELB cannot be attached to multiple subnets in the same AZ.
```

With this pull request, you can successfully run in a single az by just setting both availability zone and subnets:
```
export TF_VAR_availability_zones=us-east-1a
export TF_VAR_public_subnets=10.0.1.0/24
```

BTW, why would I want to run in a single AZ?  I'm trying to set up Flocker, which only supports a single AZ.  If anyone has gotten this working please send me a message.


wallies:
@joe1chen we have looked at flocker and had a chat to the team at clusterhq but alot of their setups for mesos and pure docker arent quite supported yet but this PR makes perfect sense

PR #692, opened 2016-04-07
Update vagrant.md

dllewellyn:


tayzlor:
while this is true, i'm not sure its the right fix. We explicitly disable hostmanager here https://github.com/Capgemini/Apollo/blob/master/Vagrantfile#L31

So I think instead of a docs update we need to wrap the above line in -

```if Vagrant.has_plugin?('vagrant-hostmanager') then```

tayzlor:
Actually - I tell a lie - we explicitly use hostmanager hereafter https://github.com/Capgemini/Apollo/blob/master/Vagrantfile#L135

PR #695, opened 2016-04-14
spark support for apollo

wallies:


tayzlor:
Code looks good from here

PR #697, opened 2016-04-15
Letsencrypt

wallies:


requirements.yml
```diff
@@ -4,3 +4,4 @@
 
 - src: defunctzombie.coreos-bootstrap
   name: coreos_bootstrap
+  when: onpremise|bool
```
tayzlor: will this work in this context? since this is for doing a `ansible-galaxy install` . Do we not need to set the condition in the playbook (site.yml) ?

roles/traefik/tasks/main.yml
```diff
@@ -18,6 +18,15 @@
   tags:
     - traefik
 
+- name: Create Lets Encrypt file
+  become: yes
+  file:
+    path: "/tmp/acme.json"
```
tayzlor: should we place this somewhere a bit more permanent? maybe in /opt/ somewhere?

roles/traefik/templates/traefik.toml.j2
```diff
@@ -1,7 +1,28 @@
...
+      certFile = "/etc/traefik/ssl/ca.pem"
+      keyFile = "/etc/traefik/ssl/ca-key.pem"
 graceTimeOut = 10
 logLevel = "{{ traefik_log_level }}"
 
+[acme]
+email = "test@traefik.io"
+storageFile = "/config/acme.json"
+entryPoint = "https"
+onDemand = trues
```
tayzlor: is this a typo here?

terraform/aws/public-cloud/main.tf
```diff
@@ -35,6 +37,16 @@ module "aws-keypair" {
   public_key_filename = "${var.public_key_file}"
 }
 
+module "ca" {
+  source            = "github.com/Capgemini/tf_tls//ca"
```
tayzlor: addtional `/` here

terraform/aws/public-cloud/main.tf
```diff
@@ -35,6 +37,16 @@ module "aws-keypair" {
   public_key_filename = "${var.public_key_file}"
 }
 
+module "ca" {
```
tayzlor: maybe rename this module to be a bit more explicit as its specifically traefik ca certs

terraform/aws/public-cloud/main.tf
```diff
@@ -1,6 +1,8 @@
 variable "access_key" {}
```
tayzlor, line 1: what about other cloud variations ? (digitalocean/aws private/etc..)

tayzlor:
How are we handling certificate renewal ?

terraform/aws/public-cloud/main.tf
wallies: this is the proper syntax as per here when using subdirectories https://www.terraform.io/docs/modules/sources.html

group_vars/all
```diff
@@ -31,6 +31,7 @@ marathon_enabled: true
 datadog_enabled: false
 scope_enabled: false
 docker_host: 'unix:///var/run/weave/weave.sock'
+onpremise: false
```
enxebre: I would just remove this flag as it's assuming too much, if it's on premise probably a different solution for every specific case need to be put in place and "defunctzombie.coreos-bootstrap" might not be always needed, or maybe just rename the flag to ansible-coreos-bootstrap or something, as per @tayzlor comment below we could move to the conditional to the playbook level

wallies:
@tayzlor pretty sure certificate renewal for the traefik ca that we generate via terraform doesnt matter. Thats just there so traefik starts up. The letsencrypt cert is a 90 day cert that has a auto renew if less than 7 days to expire.

PR #698, opened 2016-04-15
gpu support

wallies:


roles/mesos/tasks/agent.yml
```diff
@@ -9,6 +9,22 @@
...
+
+- name: add gpu support for slave
+  set_fact:
+    gpu: true
+  when: nvidio_drivers.stdout contains "nvidia0"
+  tags:
+    - mesos-agent
+
 - name: deploy mesos-agent service
   become: yes
```
enxebre: wonder if could do this as a bit of the systemd script?

wallies: how would you do this in a unit systemd file

roles/mesos/tasks/agent.yml
```diff
@@ -9,6 +9,22 @@
...
+  command: ls -al | grep -i "nvidia"
+  args:
+    chdir: /dev
+  register: nvidio_drivers
+  tags:
+    - mesos-agent
+
+- name: add gpu support for slave
+  set_fact:
+    gpu: true
```
tayzlor: does this need to be set to a default of false in /defaults/main.yml?

PR #701, opened 2016-04-19
use file with template as per docs

wallies:
https://www.terraform.io/docs/providers/template/r/file.html


terraform/aws/private-cloud/main.tf
```diff
@@ -61,7 +61,7 @@ module "elb" {
 
 # Generate an etcd URL for the cluster
```
enxebre: Cant see the specific case on the docs. This should probably be a https://www.terraform.io/docs/provisioners/null_resource.html anyway

